PR #1, opened 2019-03-15
WIP MVP

Lucus16:
- [x] Can parse all files in nixpkgs
- [ ] Pretty printing doesn't make stuff worse
- [ ] Is idempotent


nixfmt.cabal
```diff
@@ -6,10 +6,20 @@ build-type:          Simple
 
 executable nixfmt
```
kirelagin: It is not a good idea to put everything into the executable. You usually want to keep the amount of code in the executable as small as possible and put most of the functionality into the library, so that it can be reused to others, or by yourself later, or so that you can write tests.

kirelagin: (I would also recommend using `package.yaml` instead of a cabal file.)

src/Nixfmt/Lexer.hs
```diff
@@ -0,0 +1,73 @@
...
+    (PTBlockComment . splitLines . pack <$> manyTill anySingle (chunk "*/"))
+
+convertTrailing :: [ParseTrivium] -> Maybe Text
+convertTrailing = (\case
+        "" -> Nothing
+        tc -> Just $ cons ' ' tc
+    ) . intercalate " " . filter (/="") . map (\case
+        (PTLineComment lc) -> strip lc
+        (PTBlockComment [bc]) -> strip bc
+        _ -> "")
```
kirelagin: This is unreadable :).

src/Nixfmt/Lexer.hs
```diff
@@ -0,0 +1,73 @@
...
+
+convertLeading :: [ParseTrivium] -> Trivia
+convertLeading = Prelude.concat . map (\case
```
kirelagin: `concat . map = concatMap` (and the latter probably fuses better)

src/Nixfmt/Lexer.hs
```diff
@@ -0,0 +1,73 @@
...
+import           Data.Text            as Text hiding (filter, length, map, span)
+import           Nixfmt.Types         (Ann (..), Parser, Trivia, Trivium (..))
+import           Nixfmt.Util
+import           Text.Megaparsec      hiding (Token)
+import           Text.Megaparsec.Char
+
+data ParseTrivium = PTNewlines     Int
+                  | PTLineComment  Text
+                  | PTBlockComment [Text]
+                  deriving (Show)
```
kirelagin: Formatting: we always align code on fixed indent, so that it does not depend on the length of the identifiers on the previous line.

src/Nixfmt/Lexer.hs
```diff
@@ -0,0 +1,73 @@
...
+    (PTBlockComment bc) -> [BlockComment $ dropCommonIndentation bc])
+
+isTrailing :: ParseTrivium -> Bool
+isTrailing (PTLineComment _)    = True
+isTrailing (PTBlockComment [_]) = True
+isTrailing _                    = False
+
+convertTrivia :: [ParseTrivium] -> (Maybe Text, Trivia)
+convertTrivia pts = let (trailing, leading) = span isTrailing pts
+                    in (convertTrailing trailing, convertLeading leading)
```
kirelagin: Alignment.

src/Nixfmt/Types.hs
```diff
@@ -0,0 +1,242 @@
...
+data Trivium = EmptyLine
+             | LineComment     Text
+             | BlockComment    [Text]
+
+type Trivia = [Trivium]
+
+instance Show Trivium where
+    show EmptyLine         = "EmptyLine"
+    show (LineComment lc)  = " # " <> unpack lc
+    show (BlockComment bc) = "/*" <> concat (map unpack bc) <> "*/"
```
kirelagin: Oh, that’s quite a bad idea, `Show` is usually used only for debugging output, not for actual user-facing output. I’m not sure if the formatting library that you use provides some sort of typeclass for this, but if it doesn’t I personally prefer the `fmt` library and its `Buildable` class, that allows you to work directly with `Text` without all these conversions that you have to deal with. And it’s just better semantically.

src/Nixfmt/Types.hs
```diff
@@ -0,0 +1,242 @@
...
+            | InfixN
+            | InfixR
+            | Postfix
+            deriving (Show)
+
+data Operator = Op Fixity Token
+              | Apply
+              deriving (Show)
+
+operators :: [[Operator]]
```
kirelagin: My guess is that this list of lists structure gives the precedence, but it would definitely help to have this invariant documented. (In general, it would help to have more comments :)

src/Nixfmt/Util.hs
```diff
@@ -0,0 +1,52 @@
...
+import           Data.Void
+import           Text.Megaparsec hiding (empty)
+
+type Parser = Parsec Void Text
+
+someP :: (Char -> Bool) -> Parser Text
+someP = takeWhile1P Nothing
+
+manyP :: (Char -> Bool) -> Parser Text
+manyP = takeWhileP Nothing
```
kirelagin: These names are a little misleading, as the `some` and `many` functions usually do something different 🤔.

src/Nixfmt/Util.hs
```diff
@@ -0,0 +1,52 @@
...
+someText p = Text.concat <$> some p
+
+manyText :: Parser Text -> Parser Text
+manyText p = Text.concat <$> many p
+
+someList :: Parser [a] -> Parser [a]
+someList p = Prelude.concat <$> some p
+
+manyList :: Parser [a] -> Parser [a]
+manyList p = Prelude.concat <$> many p
```
kirelagin: Overall, all of these utils have confusing names and semantics. What’s the value of having them, it doesn’t seem too hard to just inline their definitions, which will, in my opinion, result in better readability.

src/Nixfmt/Util.hs
```diff
@@ -0,0 +1,52 @@
...
+
+commonPrefix :: Eq a => [a] -> [a] -> [a]
+commonPrefix (x : xs) (y : ys) | x == y = x : commonPrefix xs ys
+commonPrefix _ _               = []
+
+commonIndentation :: [String] -> String
```
kirelagin: I see `String` and I think “suspicious”.

src/Nixfmt/Util.hs
```diff
@@ -0,0 +1,52 @@
...
+commonIndentation []        = []
+commonIndentation [x]       = takeWhile isSpace x
+commonIndentation (x:"":xs) = commonIndentation (x:xs)
+commonIndentation (x:y:xs)  = commonIndentation (commonPrefix x y : xs)
+
+dropCommonIndentation :: [Text] -> [Text]
+dropCommonIndentation unstrippedLines =
+    let strippedLines = map stripEnd unstrippedLines
+        indentation = pack $ commonIndentation $ map unpack strippedLines
```
kirelagin: I see `pack . (...) . unpack` and I think “suspicious”.

src/Nixfmt/Util.hs
```diff
@@ -0,0 +1,52 @@
...
+    , dropCommonIndentation
+    ) where
+
+import           Data.Char
+import           Data.Maybe
+import           Data.Text       as Text hiding (map, takeWhile)
+import           Data.Void
+import           Text.Megaparsec hiding (empty)
+
+type Parser = Parsec Void Text
```
kirelagin: Do you really need to fix the stream type for your utils? They all seem to be generic enough.

src/Nixfmt/Pretty.hs
```diff
@@ -0,0 +1,141 @@
...
+
+instance Pretty NixToken where
+    pretty (Identifier i) = text i
+    pretty (EnvPath p)    = text ("<" <> p <> ">")
+    pretty (NixURI u)     = text u
+    pretty (NixFloat f)   = text f
+    pretty (NixInt i)     = text $ pack $ show i
+    pretty (NixText t)    = text t
+
+    pretty t              = pretty $ show t
```
kirelagin: That’s quite error-prone. You will add new constructors to `NixToken` and will definitely forget to add cases to this definition, and the compiler will not warn you because of the catch-all pattern.

src/Nixfmt/Pretty.hs
```diff
@@ -0,0 +1,141 @@
...
+import           Nixfmt.Predoc
+import           Nixfmt.Types
+import           Prelude       hiding (break)
+
+instance Pretty Trivium where
+    pretty EmptyLine            = emptyline
+    pretty (LineComment lc)     = text "#" <> pretty lc <> hardline
+    pretty (BlockComment bc)
+        = text "/*" <>
+          hcat (intersperse hardline $ map text bc) <>
```
kirelagin: In terms of formatting, the alignment, I’m not sure if it is in our style-guide yet, but we (I) typically prefer operator-first formatting.
```suggestion
       <> hcat (intersperse hardline $ map text bc)
```

kirelagin: (Ugh, that’s just an example with one line – apparently GitHub does not support multi-line suggestions.)

src/Nixfmt/Predoc.hs
```diff
@@ -0,0 +1,195 @@
...
+isLine (Line _) = True
+isLine _        = False
+
+spanEnd :: (a -> Bool) -> [a] -> ([a], [a])
+spanEnd p = fmap reverse . span p . reverse
+
+-- | Lines before a Nest should be moved inside, breaks at the end of a nest
+-- should be moved outside. Lines at the start or end of a group should be
+-- moved outside.
+
```
kirelagin: 👾

src/Nixfmt/Predoc.hs
```diff
@@ -0,0 +1,195 @@
...
+space = Leaf (Line Space)
+
+hardline :: Doc
+hardline = Leaf (Line Hardline)
+
+emptyline :: Doc
+emptyline = Leaf (Line Emptyline)
+
+hcat :: [Doc] -> Doc
+hcat = mconcat
```
kirelagin: 😕

src/Nixfmt/Predoc.hs
```diff
@@ -0,0 +1,195 @@
...
+data Predoc f
+    = Trivia Text
+    | Text Text
+    | Line Line
+    -- | Group predoc indicates either all or none of the spaces in predoc
+    -- and not in a subgroup should be converted to line breaks.
+    | Group (f (Predoc f))
+    -- | Nest n predoc indicates all line start in predoc should be indented by
+    -- n more spaces than the surroundings.
+    | Nest Int (f (Predoc f))
```
kirelagin: Here the alignment is correct. In the definition right above it is not.

src/Nixfmt/Pretty.hs
```diff
@@ -0,0 +1,141 @@
...
+        in group $ pretty parameter <> pretty colon <> space <> pretty expression
+
+    pretty (Node ContextParameter children0) = hcat $ map pretty children0
+
+    pretty (Node SetParameter children0) =
+        let Just (leading, braceOpen, children1) = splitT TBraceOpen children0
+            Just (params, braceClose, trailing) = splitT TBraceClose children1
+            go xs = case splitT TComma xs of
+                         Just (before, comma, after)
+                                 -> nest 2 (pretty before) <> break <>
```
kirelagin: > `break`

It would be lovely if, instead of hiding standard well-known definitions, you imported and used your custom combinators qualified.

src/Nixfmt/Pretty.hs
```diff
@@ -0,0 +1,141 @@
...
+instance Pretty NixAST where
+    pretty (Leaf l Nothing) = pretty l
+    pretty (Leaf l (Just comment)) = pretty l <> text (" #" <> comment) <> hardline
+    pretty (Trivia t) = pretty t
+
+    pretty (Node File children) = pretty children <> hardline
+
+    pretty (Node List children0) =
+        let Just (leading, brackOpen, children1) = splitT TBrackOpen children0
+            Just (items, brackClose, trailing) = splitT TBrackClose children1
```
kirelagin: Wow-wow, spooky!
`-Wincomplete-uni-patterns`

src/Nixfmt/Util.hs
Lucus16: To be honest, I feel these names are very accurate. They are variants of `some` and `many` with some additional convenience. In fact, I feel like the `takeWhileP` names stand out for doing something very similar, but being named completely different.

src/Nixfmt/Predoc.hs
Lucus16: This is just to match the interface of the `prettyprinter` library I'm wrapping.

src/Nixfmt/Predoc.hs
Lucus16: I improved the documentation, let me know if it's still unclear.

src/Nixfmt/Pretty.hs
Lucus16: I wish a formatter would do this for me :p. I encounter a lot of cases in Haskell where I don't know what the right way is to format something. I'll fix this when I'm too tired to do more useful stuff sometime. Can you give me the reasoning for putting operators first, since it may apply to nixfmt as well?

Lucus16:
This should have taken care of most of your feedback.

default.nix
```diff
@@ -0,0 +1,3 @@
+{ pkgs ? import <nixpkgs> {} }:
```
yorickvP: callCabal2nix?

shell.nix
```diff
@@ -2,15 +2,14 @@
 
 let
   inherit (pkgs) stdenv;
-  hpkgs = pkgs.haskell.packages.ghc863;
-  ghc = hpkgs.ghcWithHoogle (hpkgs: with hpkgs; [
+  ghc = pkgs.haskellPackages.ghcWithHoogle (hpkgs: with hpkgs; [
```
yorickvP: is it possible to use `(import ./default.nix).env` here instead of repeating the packages?

src/Nixfmt/Parser.hs
```diff
@@ -0,0 +1,266 @@
...
+import           Nixfmt.Types
+import           Nixfmt.Util
+
+-- HELPER FUNCTIONS
+
+ann :: (a -> b) -> Parser a -> Parser (Ann b)
+ann f p = try $ lexeme $ f <$> p
+
+pair :: Parser a -> Parser b -> Parser (a, b)
+pair p q = (,) <$> p <*> q
```
yorickvP:
```suggestion
pair p q = liftM2 (,)
```

src/Nixfmt/Predoc.hs
```diff
@@ -0,0 +1,237 @@
...
+
+moveLinesOut (Nest level xs : ys) =
+    let movedOut     = moveLinesOut xs
+        (pre, rest)  = span isLine movedOut
+        (post, body) = spanEnd isLine rest
+    in pre ++ (Nest level body : (post ++ moveLinesOut ys))
+
+moveLinesOut (x : xs) = x : moveLinesOut xs
+
+dropEmpty :: DocList -> DocList
```
yorickvP: sounds like you should implement/derive `Foldable` for `Doc`

PR #6, opened 2019-05-13
Repo review

Martoon-00:
I dunno how to comment on code already existing in the repo,
so here is a commit which gathers all the code.


shadow/src/Nixfmt/Parser.hs
```diff
@@ -0,0 +1,317 @@
...
+stripFirstLine (x : xs)
+    | isEmptyLine x = xs
+    | otherwise     = x : xs
+
+partsInit :: [StringPart] -> [Text]
+partsInit line@(TextPart t : _)
+    | isEmptyLine line              = []
+    | otherwise                     = [t]
+partsInit (Interpolation _ _ _ : _) = [""]
+partsInit []                        = []
```
Martoon-00: It takes head, why did you call it `partsInit`?

shadow/src/Nixfmt/Lexer.hs
```diff
@@ -0,0 +1,97 @@
...
+
+splitLines :: Text -> [Text]
+splitLines = dropWhile Text.null . dropWhileEnd Text.null
+    . map Text.stripEnd . Text.lines . replace "\r\n" "\n"
+
+stripIndentation :: Int -> Text -> Text
+stripIndentation n t = fromMaybe (stripStart t) $ stripPrefix (Text.replicate n " ") t
+
+commonIndentation :: Int -> [Text] -> Int
+commonIndentation def = foldr min def . map (Text.length . Text.takeWhile (==' '))
```
Martoon-00: Minor: maybe it's only me, but it seems sad to have two functions with the same name, especially since they are not used only locally in the modules they are defined. Why not call this function e.g. `commonIndentationLength`?

shadow/main/Main.hs
```diff
@@ -0,0 +1,83 @@
...
+import Control.Concurrent
+import Control.Concurrent.ParallelIO.Local
+import Data.Either
+import qualified Data.Text.IO as TextIO
+import GHC.Conc (numCapabilities)
+import System.Console.CmdArgs
+import System.Exit
+import System.IO
+import System.Posix.Process (exitImmediately)
+import System.Posix.Signals
```
Martoon-00: You have to specify what do you import [explicitly](https://github.com/serokell/serokell-util/blob/master/serokell-style.md#imports) (unless you define your own style-guide).

shadow/src/Nixfmt/Parser.hs
```diff
@@ -0,0 +1,317 @@
...
+    ]
+
+reserved :: Token -> Parser (Ann Token)
+reserved t = try $ lexeme $ rawSymbol t
+    <* lookAhead (satisfy (\x -> not $ identChar x || pathChar x))
+
+-- VALUES
+
+integer :: Parser (Ann Token)
+integer = ann Integer L.decimal
```
Martoon-00: Doesn't nix allow negative numbers?

shadow/main/Main.hs
gromakovsky: Btw, our default style guide is [here](https://github.com/serokell/style).

shadow/main/Main.hs
```diff
@@ -0,0 +1,83 @@
...
+    writeChan errChan Nothing
+    return results
+
+main :: IO ()
+main = do
+    _ <- installHandler keyboardSignal
+            (Catch (exitImmediately $ ExitFailure 2)) Nothing
+    opts <- cmdArgs options
+    results <- formatParallel $ case files opts of
+                   [] -> pure $ formatStdio (width opts)
```
gromakovsky: I think `[formatStdio (width opts)]` would be slightly more understandable.

shadow/src/Nixfmt/Pretty.hs
```diff
@@ -0,0 +1,362 @@
...
+        <> pretty paropen <> pretty trailing <> line
+        <> nest 2 (sepBy hardline binders') <> line
+        <> pretty parclose
+    where binders' = case leading of
+                          [] -> binders
+                          _  -> BinderTrivia leading : binders
+
+prettyTerm (String x) = pretty x
+
+prettyTerm _ = undefined
```
gromakovsky: It's better to use `error` and write at least something that can give a hint where a bug happened if it even happens. Also consider adding `HasCallStack` constraint to this function.

shadow/src/Nixfmt/Parser.hs
Lucus16: It does, but they are simply parsed as a negation expression wrapping a positive number.

shadow/main/Main.hs
```diff
@@ -0,0 +1,83 @@
...
+    } &= help "Format Nix source code"
+
+formatStdio :: Int -> IO Result
+formatStdio w = do
+    contents <- TextIO.getContents
+    mapM TextIO.putStr $ format w "<stdin>" contents
+
+formatFile :: Int -> FilePath -> IO Result
+formatFile w path = do
+    contents <- TextIO.readFile path
```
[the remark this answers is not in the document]
Lucus16: To be honest, I have no idea how to handle encoding properly, but forcing UTF-8 everywhere seems too easy. I grepped through the Nix manual and source code for a bit and I don't think it forces UTF-8, so people might have nix files in other encodings.

Lucus16:
> I dunno how to comment on code already existing in the repo,
> so here is a commit which gathers all the code.

I'm going to change this PR a bit so I can actually push code to it that can be merged.

shadow/main/Main.hs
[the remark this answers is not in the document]
Lucus16: Made a separate issues for this, see #8.

kirelagin:
@Martoon-00 hiii

Lucus16:
This is probably no longer relevant.

PR #12, opened 2019-05-26
Fix handling of UTF-8 with bad locales

### LINKED ISSUE #8: Handling text encoding
> Is encoding of .nix file specified anywhere? For example, if it's UTF-8, this call may fail if C locale is set. Please double check. The same applies to formatStdio (there also TextIO.putStr probably can fail if C locale is set).

We should support other encodings, but we should also avoid failing on UTF-8 files when a different locale is set. Maybe we should try both when reading? I don't know how this is usually handled.

### LINKED ISSUE #11: Keyboard interrupt does not clean up temporary files
Currently, a keyboard interrupt is handled by immediately exiting. The temporary files created by `atomicWriteFile` are not cleaned up because of that. I have yet to figure out if the `atomic-write` library includes error handling to delete the file if it exits more gracefully.

kirelagin:
Fixes #8.
Fixes #11.

This PR adds a helper module (which is intended to eventually become a separate library) and changes the `nixfmt` binary to use functions from this new module for IO.

Specifically:

* If std{in,out,err} is terminal and the locale does not support UTF-8, it will transliterate (well, more like, replace with question marks) unsupported characters. The former two handles are used in the stdin/out processing mode, the latter is used for error reporting (errors can potentially include weird characters too).
* If std{in,out,err} is file, always assume it is in UTF-8.
* In the file processing mode, always assume files are in UTF-8.

## Test cases

```nix
# test.nix
let
 foo = "привет";
  in
   foo
```

These did not work before, now they do:
* `LANG=C nixfmt < test.nix`
* `LANG=C nixfmt < test.nix > out.nix`
* `LANG=C nixfmt text.nix`


kirelagin:
@Lucus16 Let me know what you think about the atomic thing. How should we proceed? I would suggest we drop the library.

@gromakovsky @Martoon-00 Guys, pls, review, since you are now experts :).

kirelagin:
@Lucus16 Of all the atomic write libraries I found on Hackage, this one is the only one that provides a sensible API: http://hackage.haskell.org/package/safeio-0.0.5.0/docs/System-IO-SafeWrite.html#v:withOutputFile

However, as far as I can tell, it does not copy any attributes, not even the mode. Even worse, it uses a system function to create the temporary file, and this function sets the mode to a very restricted one (only the owner can read). But the author got the interface right, so maybe we should contribute to it, or fork it, drop conduit, and make it preserve the attributes?

kirelagin:
Ok, anyway, I reimplemented atomic write myself 🙄.

src/System/IO/Atomic.hs
```diff
@@ -0,0 +1,49 @@
...
+import Control.Monad (when)
+import System.Directory (doesFileExist, removeFile, renameFile)
+import System.FilePath (takeDirectory, takeFileName)
+import System.IO (Handle, hClose, openTempFileWithDefaultPermissions)
+import System.Posix.Files (fileMode, getFileStatus, setFileMode)
+
+
+withOutputFile ::
+     FilePath  -- ^ Final file path
+  -> (Handle -> IO a)  -- ^ IO action that write to the file
```
gromakovsky: s/write/writes

src/System/IO/Atomic.hs
```diff
@@ -0,0 +1,49 @@
+{- © 2019 Serokell <hi@serokell.io>
+ -
+ - SPDX-License-Identifier: MPL-2.0
+ -}
+
+-- | Helpers for atomic file IO.
+--
+-- This module is a crossover of the `atomic-write` and `safeio` libraries
```
Lucus16: I'd much rather you give a description of what this module does than where it comes from, since that's what I'd want to know if I were reading the documentation.

main/Main.hs
```diff
@@ -40,8 +43,12 @@ formatStdio w = do
 
 formatFile :: Int -> FilePath -> IO Result
 formatFile w path = do
-    contents <- TextIO.readFile path
-    mapM (atomicWriteFile path) $ format w path contents
+    contents <- readFileUtf8 path
+    mapM atomicWriteFile $ format w path contents
+  where
+    atomicWriteFile t = withOutputFile path $ \h -> do
+      hSetEncoding h utf8
```
Lucus16: Is it possible to make the two new modules more composable so no implementation details leak here?

src/System/IO/Atomic.hs
kirelagin: > Helpers for atomic file IO.

Not good enough? 🤔

main/Main.hs
kirelagin: I don’t think so, and I also don’t think any implementation details leak here. Correct handling of the data encoding is ultimately the responsibility of the programmer, the utf8 module just has helpers for frequently used tasks that can be considered “best practices”.

kirelagin:
> Would be good to add tests to CI.
Indeed.

src/System/IO/Atomic.hs
Lucus16: > This module provides functions that ensure a file is written atomically by first writing the new contents to a temporary file. It also ensures this temporary file is cleaned up when an exception is raised.
Something like that.

main/Main.hs
Lucus16: Okay, it can't be helped then.

Lucus16:
The way I see it, the library should handle formatting and the executable should handle the CLI and filesystem interface.

kirelagin:
Well, these two are supposed to be two separate libraries eventually anyway, sooo... I don’t think it matters at all, as long as we do not keep them here for too long. But feel free to move the to the executable if you think it’s better this way.

kirelagin:
RE dependencies: I don’t think I introduced any dependencies that are not boot libraries.

main/System/IO/Atomic.hs
```diff
@@ -0,0 +1,64 @@
...
+    copyAttributes (tpath, _th) = do
+      exists <- doesFileExist path
+      when exists $ do
+        getFileStatus path >>= setFileMode tpath . fileMode
+
+    commit :: (FilePath, Handle) -> IO ()
+    commit (tpath, th) = hClose th *> renameFile tpath path
+
+    rollback :: (FilePath, Handle) -> IO ()
+    rollback (tpath, th) = hClose th *> removeFile tpath
```
Martoon-00: If `hClose` fails for some reason, full cleanup won't occur.

main/System/IO/Atomic.hs
```diff
@@ -0,0 +1,64 @@
...
+-- it is Unix-specific and there is definitely room for improvement even on Unix.
+withOutputFile ::
+     FilePath  -- ^ Final file path
+  -> (Handle -> IO a)  -- ^ IO action that writes to the file
+  -> IO a
+withOutputFile path act = bracketOnError begin rollback $ \(tpath, th) -> do
+    copyAttributes (tpath, th)
+    result <- act th
+    commit (tpath, th)
+    pure result
```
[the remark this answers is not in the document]
kirelagin: ouch

main/System/IO/Atomic.hs
kirelagin: Hmmm riiiight. I guess it shouldn’t matter in what order we do these two things, so it’s better to remove the file first, and close the handle after that?

main/System/IO/Atomic.hs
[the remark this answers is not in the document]
kirelagin: Or should I just implement my own bracket that masks both `commit` and `rollback`? What do you think?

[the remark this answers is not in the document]
Martoon-00: Looks like less trivial but a more proper solution.

main/System/IO/Atomic.hs
Martoon-00: Wow, if it works then ok.

But can't `removeFile` fail as well? (probably in a very unlucky case like someone `rm -rf` the directory, but still).

kirelagin: Ok, I am now not sure about this one. If `hClose` fails then we are probably in really deep trouble, and thus there is probably no need to try to recover or clean up or anything, we are screwed. 🤔

kirelagin: Well, yes, `removeFile` can fails of course, e.g. if someone, as you say, removed this file before us, but I would speculate, that this is a condition worth reporting (i.e. we should let this exception propagate).

kirelagin:
Oh, actually, it fixes #11 as well, what a coincidence!

main/System/IO/Atomic.hs
Martoon-00: Okay, let's leave it as is.

Lucus16:
It's probably best to drop any exceptions in each of the parts of `rollback` since a more important exception is already being thrown.

kirelagin:
> It's probably best to drop any exceptions in each of the parts of `rollback` since a more important exception is already being thrown.

Isn’t it how `onException` works?

Lucus16:
What I mean that exceptions for each of the parts must be dropped since neither should prevent the other from being executed afaict.

kirelagin:
No, I don’t think so 🤔. As I said in some comment thread, it is really weird if `hClose` fails, it means we are in a very-very serious trouble, so it’s probably better to stop trying. Also, I don’t think we want to drop those exceptions: as a programmer, I would prefer to know that my rollback action failed.

kirelagin:
Oh, and, on a separate note, fuck my life, if I am reading it correctly, `onException` from `base` will actually rethrow the second exception, if it happens, not the first one. `onException` in `safe-exceptions` rethrows the first one in any case, though.

Martoon-00:
Yeah, also find it weird. And `bracket` seem to inherit that behavior

kirelagin:
Rebased.

kirelagin:
Signed.

PR #21, opened 2019-07-16
[INT-82] Format expressions

Lucus16:
This should solve the main issue with indentation. There are still many edge cases that need to be addressed one by one, but most code should now get formatted nicely.

The first four commits are just refactoring and should not have semantic changes. The final commit is what solves the indentation issue.

### DIFF
--- a/src/Nixfmt/Predoc.hs
+++ b/src/Nixfmt/Predoc.hs
@@ -26,7 +26,6 @@ module Nixfmt.Predoc
     , Doc
     , Pretty
     , pretty
-    , flatten
     , fixup
     , layout
     , textWidth
@@ -50,29 +49,30 @@ data Spacing
     deriving (Show, Eq, Ord)
 
 data DocAnn
-    = NoAnn
     -- | Node Group docs indicates either all or none of the Spaces and Breaks
     -- in docs should be converted to line breaks.
-    | Group
+    = Group
     -- | Node (Nest n) docs indicates all line start in docs should be indented
     -- by n more spaces than the surroundings.
     | Nest Int
     deriving (Show, Eq)
 
-data Doc
+data DocE
     = Text Text
     | Spacing Spacing
-    | Node DocAnn [Doc]
+    | Node DocAnn Doc
     deriving (Show, Eq)
+
+type Doc = [DocE]
 
 class Pretty a where
     pretty :: a -> Doc
 
 instance Pretty Text where
-    pretty = Text
+    pretty = pure . Text
 
 instance Pretty String where
-    pretty = Text . pack
+    pretty = pure . Text . pack
 
 instance Pretty Doc where
     pretty = id
@@ -81,46 +81,39 @@ instance Pretty a => Pretty (Maybe a) where
     pretty Nothing  = mempty
     pretty (Just x) = pretty x
 
-instance Semigroup Doc where
-    x <> (Node NoAnn xs) = Node NoAnn (x:xs)
-    left <> right = Node NoAnn [left, right]
-
-instance Monoid Doc where
-    mempty = Node NoAnn []
-    mconcat = Node NoAnn
-
 text :: Text -> Doc
-text = Text
+text "" = []
+text t  = [Text t]
 
 group :: Pretty a => a -> Doc
-group = Node Group . pure . pretty
+group = pure . Node Group . pretty
 
 nest :: Int -> Doc -> Doc
-nest level = Node (Nest level) . pure
+nest level = pure . Node (Nest level)
 
 softline' :: Doc
-softline' = Spacing Softbreak
+softline' = [Spacing Softbreak]
 
 line' :: Doc
-line' = Spacing Break
+line' = [Spacing Break]
 
 softline :: Doc
-softline = Spacing Softspace
+softline = [Spacing Softspace]
 
 line :: Doc
-line = Spacing Space
+line = [Spacing Space]
 
 hardspace :: Doc
-hardspace = Spacing Hardspace
+hardspace = [Spacing Hardspace]
 
 hardline :: Doc
-hardline = Spacing Hardline
+hardline = [Spacing Hardline]
 
 emptyline :: Doc
-emptyline = Spacing Emptyline
+emptyline = [Spacing Emptyline]
 
 newline :: Doc
-newline = Spacing (Newlines 1)
+newline = [Spacing (Newlines 1)]
 
 sepBy :: Pretty a => Doc -> [a] -> Doc
 sepBy separator = mconcat . intersperse separator . map pretty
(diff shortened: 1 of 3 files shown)


src/Nixfmt/Predoc.hs
```diff
@@ -89,44 +81,46 @@ instance Pretty a => Pretty (Maybe a) where
...
-    left <> right = Node left right
+instance Semigroup Doc where
+    x <> (Node NoAnn xs) = Node NoAnn (x:xs)
+    left <> right = Node NoAnn [left, right]
 
-instance Monoid (Tree a) where
-    mempty = EmptyTree
+instance Monoid Doc where
+    mempty = Node NoAnn []
+    mconcat = Node NoAnn
```
gromakovsky: 1. This `mconcat` differs from `<>` because it doesn't consider `Node NoAnn xs` special case. Even if it was the same, it's better to write `mconcat = (<>)` to make it clear.
2. Even if you write `mconcat = (<>)` it won't be a monoid, because `mempty <> x = Node NoAnn [mempty, x] ≠ x` for all `x ≠ mempty`. So you should also explicitly consider `Node NoAnn xs` case in the `Semigroup` instance definiton.

src/Nixfmt/Predoc.hs
```diff
@@ -89,44 +81,46 @@ instance Pretty a => Pretty (Maybe a) where
     pretty Nothing  = mempty
     pretty (Just x) = pretty x
 
-instance Semigroup (Tree a) where
-    left <> right = Node left right
+instance Semigroup Doc where
+    x <> (Node NoAnn xs) = Node NoAnn (x:xs)
```
gromakovsky: Brackets are redundant here, btw.

src/Nixfmt/Predoc.hs
```diff
@@ -135,49 +129,42 @@ sepBy separator = mconcat . intersperse separator . map pretty
...
+              case x of
+                   Node NoAnn ys  -> go done (ys ++ todo)
+                   Node ann   ys  -> go (Node ann (go [] ys) : done) todo
+                   Text ""        -> go done todo
+                   l              -> go (l : done) todo
+
+isSpacing :: Doc -> Bool
 isSpacing (Spacing _) = True
-isSpacing _           = False
+isSpacing _                  = False
```
gromakovsky: Now this indentation looks too weird.

src/Nixfmt/Predoc.hs
```diff
@@ -192,32 +179,28 @@ mergeSpacings Hardspace    (Newlines x) = Newlines x
...
+moveLinesIn (Spacing l : Node (Nest level) xs : ys) =
+    Node (Nest level) (Spacing l : moveLinesIn xs) : moveLinesIn ys
 
-moveLinesIn (Group xs : ys) =
-    Group (moveLinesIn xs) : moveLinesIn ys
+moveLinesIn (Node ann xs : ys) =
+    Node ann (moveLinesIn xs) : moveLinesIn ys
 
-moveLinesIn (x : xs) = x : moveLinesIn xs
+moveLinesIn (x : xs)                     = x : moveLinesIn xs
```
gromakovsky: Something looks wrong with indentation here.

src/Nixfmt/Predoc.hs
```diff
@@ -129,35 +122,24 @@ sepBy separator = mconcat . intersperse separator . map pretty
...
-isSpacing :: Doc -> Bool
+isSpacing :: DocE -> Bool
 isSpacing (Spacing _) = True
 isSpacing _                  = False
 
 spanEnd :: (a -> Bool) -> [a] -> ([a], [a])
 spanEnd p = fmap reverse . span p . reverse
 
--- | Fix up a Doc in multiple stages:
+-- | Fix up a DocE in multiple stages:
```
gromakovsky: AFAIU, it should still be `Doc`, not `DocE`.

src/Nixfmt/Predoc.hs
```diff
@@ -50,29 +49,30 @@ data Spacing
...
     -- in docs should be converted to line breaks.
-    | Group
+    = Group
     -- | Node (Nest n) docs indicates all line start in docs should be indented
     -- by n more spaces than the surroundings.
     | Nest Int
     deriving (Show, Eq)
 
-data Doc
+data DocE
```
gromakovsky: What does this `E` stand for? Element? Would be good to write a comment.

gromakovsky:
I haven't reviewed last commit yet, will do it now.
In general I must say that it's hard to review actual formatting logic for a person not familiar with this code (or just barely familiar, who saw it once, but doesn't work with it). If there were tests, it would give some confidence that code actually works. Unfortunately, I couldn't find any tests here. I recommend adding tests ASAP.
I will check that code is generally sane and reasonable, but if there is some logical issue, I probably won't catch it.

src/Nixfmt/Pretty.hs
```diff
@@ -103,7 +103,7 @@ prettyTerm (List (Ann paropen Nothing []) [ListItem item] parclose)
         = pretty paropen <> pretty item <> pretty parclose
 
 prettyTerm (List (Ann paropen trailing leading) items parclose)
-    = pretty paropen <> pretty trailing <> line
+    = nest 0 $ pretty paropen <> pretty trailing <> line
```
gromakovsky: What is the purpose of `nest 0`?

Lucus16: I'll add a comment to the definition of `nest` to explain this.

Lucus16:
Thanks for catching my mistakes! I'll improve the documentation a little. As for tests, it's probably about time I start adding some, but so far, the easiest way to test is to run nixfmt on a clean checkout of `nixpkgs` and see if the resulting diff improves formatting. Run from the nixpkgs checkout directory, takes about 30 seconds on my laptop:

    time find -name '*.nix' | xargs -P8 /path/to/nixfmt

src/Nixfmt/Predoc.hs
```diff
@@ -55,32 +48,31 @@ data Spacing
...
+    deriving (Show, Eq)
 
-deriving instance Eq (Predoc Tree)
-deriving instance Eq (Predoc [])
-deriving instance Show (Predoc Tree)
-deriving instance Show (Predoc [])
-
-type Doc = Tree (Predoc Tree)
-type DocList = [Predoc []]
+type Doc = [DocE]
```
Martoon-00: I'm not sure how much files this tool is expected to work with, and probably this concern has already been raised, but have you considered replacing `[]` with something else? E.g. on each nested code block you perform concatenation of documentation items and it takes linear time.

Usually, for an incremental building they use data structures which provide lazy concatenation with `O(1)` amortized complexity. For instance, `DList` can be picked; although, it would be slightly less convenient to use.

Lucus16: Actually, I used a custom data structure before as this diff shows and it made the code significantly more complicated. From what I can tell, using lists is less than 20% slower on `nixpkgs`, so this will do for now. I can look at optimization in the future, but I want it to be guided by profiling.

PR #27, opened 2019-07-25
Create CODE_OF_CONDUCT.md

Lucus16:


yorickvP:
Let's do this at a later stage, when there is a community. We'll have to think more about which CoC to pick.

jagajaga:
👎

Issue #32, opened 2019-07-29
Move comments to the syntax following them


Lucus16:
Currently, all comments are parsed after tokens, but comments generally apply to the token that follows it, with the exception of trailing comments on the same line. Many formatting cases based on reordering expressions can't be done properly until comments are attached to the tokens they apply to. I tried parsing comments before the tokens, but this cannot be done efficiently and elegantly.

The best way forward is to make a pass over the entire syntax tree and moving comments from after their intial token to before the next token. This pass can simultaneously move some comments into better positions where they don't interrupt formatting flow.

PR #33, opened 2019-07-29
Quick fix for commas not having spacing

### LINKED ISSUE #31: Default arguments and line breaks
### Description
Avoid adding line breaks before function parameters and corresponding default values.

### Small example input

I saw an announcement regarding this tool, and the first example I took a look at was [emacs](https://github.com/serokell/nixpkgs/blob/formatted/pkgs/applications/editors/emacs/default.nix). I was not expecting the increasing indentation of function parameters, and the line breaks after the `?` characters makes reading the default values more challenging.

### Expected output

Treat function parameters specially by not inserting line breaks between parameter names and default values. 

Wrap parameter lists at commas.

Do not increase the indentation level within a parameter list.

Lucus16:
Fixes #31. Handling of set abstractions can still be improved by quite a bit, but it would be messy until #32 is fixed.

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -20,9 +20,9 @@ import Nixfmt.Predoc
   (Doc, Pretty, emptyline, group, hardline, hardspace, hcat, line, line', nest,
   newline, pretty, sepBy, softline, softline', text, textWidth)
 import Nixfmt.Types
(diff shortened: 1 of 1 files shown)


kirelagin:
@yorickvP Release?

Lucus16:
Done. (v0.2.1)

PR #39, opened 2019-08-07
Add check and quiet flags

### LINKED ISSUE #38: Check functionality for CI
It would be nice if `nixfmt` had a `--check` flag that exited with 0 if the file is correctly formatted and exited with 1 (and error messages on `stderr`) otherwise.

That way, the tool could be used to verify formatting during CI.

Lucus16:
Fixes #38

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -9,9 +9,12 @@
 module Main where
 
 import Control.Concurrent (Chan, forkIO, newChan, readChan, writeChan)
+import Data.Bifunctor (first)
 import Data.Either (lefts)
+import Data.Text (Text)
 import GHC.IO.Encoding (utf8)
-import System.Console.CmdArgs (Data, Typeable, args, cmdArgs, help, typ, (&=))
+import System.Console.CmdArgs
+  (Data, Typeable, args, cmdArgs, help, summary, typ, (&=))
 import System.Exit (ExitCode(..), exitFailure, exitSuccess)
 import System.IO (hPutStr, hSetEncoding, stderr)
 import System.Posix.Process (exitImmediately)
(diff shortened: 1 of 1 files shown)


main/Main.hs
```diff
@@ -23,59 +26,95 @@ import Nixfmt
...
-    mapM atomicWriteFile $ format w path contents
+    , width = 80 &= help "Maximum width in characters"
+    , check = False &= help "Check whether files are formatted"
+    , quiet = False &= help "Do not report errors"
+    } &= summary "Format Nix source code"
+
+format' :: Int -> FilePath -> Text -> Either String Text
+format' w path = first errorBundlePretty . format w path
+
+data Target = Target (IO Text) FilePath (Text -> IO ())
```
kirelagin: IMHO, it would be better (for self-documentation) to use a record here.
```suggestion
data Target = Target
  { tDoRead :: IO Text
  , tPath :: FilePath
  , tDoWrite :: Text -> IO ()
  }
```

And then instead of `(Target doRead path _)` you’d be able to do `Target{tDoRead, tPath}`.

main/Main.hs
```diff
@@ -23,59 +26,95 @@ import Nixfmt
...
-formatFile :: Int -> FilePath -> IO Result
-formatFile w path = do
-    contents <- readFileUtf8 path
-    mapM atomicWriteFile $ format w path contents
+    , width = 80 &= help "Maximum width in characters"
+    , check = False &= help "Check whether files are formatted"
+    , quiet = False &= help "Do not report errors"
+    } &= summary "Format Nix source code"
+
+format' :: Int -> FilePath -> Text -> Either String Text
```
kirelagin: IMHO, this `Int` everywhere is a little confusing, it could be a `type Width = Int` instead 🙄.

main/Main.hs
```diff
@@ -23,59 +26,95 @@ import Nixfmt
...
+format' w path = first errorBundlePretty . format w path
+
+data Target = Target (IO Text) FilePath (Text -> IO ())
+
+formatTarget :: Int -> Target -> IO Result
+formatTarget w (Target doRead path doWrite)
+    = doRead >>= mapM doWrite . format' w path
```
kirelagin: It’s very uncool when the data moves in two different directions within a single line. Try:
```suggestion
    = doRead >>= format' w path & mapM doWrite
```
(I’m not 100% if the precedences are right here, but, I think, they are.)

Lucus16: This did not typecheck, even after parenthesizing, so I used do-notation in the end.

main/Main.hs
```diff
@@ -43,25 +44,31 @@ options = Nixfmt
...
+    , tDoWrite :: Text -> IO ()
+    }
 
-formatTarget :: Int -> Target -> IO Result
-formatTarget w (Target doRead path doWrite)
-    = doRead >>= mapM doWrite . format' w path
+formatTarget :: Width -> Target -> IO Result
+formatTarget w Target{tDoRead, tPath, tDoWrite} = do
+    contents <- tDoRead
+    formatted <- return $ format' w tPath contents
```
gromakovsky: `a <- return b` is `let a = b`.

main/Main.hs
```diff
@@ -23,59 +26,102 @@ import Nixfmt
...
+data Target = Target
+    { tDoRead :: IO Text
+    , tPath :: FilePath
+    , tDoWrite :: Text -> IO ()
+    }
+
+formatTarget :: Width -> Target -> IO Result
+formatTarget w Target{tDoRead, tPath, tDoWrite} = do
+    contents <- tDoRead
+    formatted <- return $ format' w tPath contents
```
kirelagin:
```suggestion
    let formatted = format' w tPath contents
```

PR #40, opened 2019-08-09
Comply with REUSE 3.0

kirelagin:
* REUSE 3.0 no longer allows `Valid-License-Identifier`
* Add missing licensing information to some files


Lucus16:
Can the `LICENSE` file be dropped? I prefer to keep the root of the repository clean, I think just the `LICENSES` directory should be enough.

kirelagin:
It can be dropped in the sense that it will not affect REUSE 3.0 compliance, but, unfortunately, this file is used by GitHub to detect the license. If we don’t care about the right license being displayed in the GUI, then ok.

Lucus16:
Ok, sucks, but I guess we'll keep it.

Lucus16:
Patched missing year.

PR #57, opened 2020-02-14
Ignore symlinks when formatting

### LINKED ISSUE #56: Nixfmt does overwrite symlinks with reformatted contents
When I have a symlink and pass it to `nixfmt` (eg because of using globbing from the shell as in `nixfmt **/*.nix`) it gets overwritten by a regular file containing the formatted target.

`nixfmt` should either ignore symlinks or write to the symlinks target, rather than overwriting the symlink.

```plain
$ readlink -f default.nix
/home/nmelzer/.config/nixpkgs/hosts/vbox-arch.nix
$ nixfmt default.nix
$ readlink -f default.nix
/home/nmelzer/.config/nixpkgs/hosts/default.nix
$ nixfmt --version
Format Nix source code
```

Sadly it seems as if in my version of `nixfmt` #50 isn't fixed yet.

Lucus16:
Before this, symlinks were replaced with a formatted file rather than
formatting the file behind the symlink. The choice was between ignoring
symlinks and following them. Following symlinks means potentially
formatting files in an unexpected location which seems much worse than
needing to wrap the arguments in a `readlink -f` when symlinks are
otherwise ignored.

Fixes #56.

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -9,6 +9,7 @@
 module Main where
 
 import Control.Concurrent (Chan, forkIO, newChan, readChan, writeChan)
+import Control.Exception (IOException, try)
 import Data.Bifunctor (first)
 import Data.Either (lefts)
 import Data.Text (Text)
@@ -17,6 +18,7 @@ import GHC.IO.Encoding (utf8)
 import Paths_nixfmt (version)
 import System.Console.CmdArgs
   (Data, Typeable, args, cmdArgs, help, summary, typ, (&=))
+import System.Directory (pathIsSymbolicLink)
 import System.Exit (ExitCode(..), exitFailure, exitSuccess)
 import System.IO (hPutStrLn, hSetEncoding, stderr)
 import System.Posix.Process (exitImmediately)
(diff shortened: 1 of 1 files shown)


Lucus16:
Ah, whoops, I broke `nixfmt` without arguments.

Lucus16:
That should fix that.

main/Main.hs
```diff
@@ -9,6 +9,7 @@
 module Main where
 
 import Control.Concurrent (Chan, forkIO, newChan, readChan, writeChan)
+import Control.Exception (IOException, try)
```
kirelagin: Please, only ever use `Control.Exception.Safe` from `safe-exceptions`.

main/Main.hs
```diff
@@ -50,42 +52,53 @@ options = Nixfmt
...
+targetName (FileTarget path) = path
+
+readTarget :: Target -> IO Text
+readTarget StdioTarget = TextIO.getContents
+readTarget (FileTarget path) = readFileUtf8 path
+
+filterTarget :: (Target -> IO Result) -> Target -> IO Result
+filterTarget operation StdioTarget = operation StdioTarget
+filterTarget operation (FileTarget path) = do
+    isSymlink <- try $ pathIsSymbolicLink path
```
kirelagin: This will also catch async exceptions which you don’t want (this probably doesn’t matter in this specific case, but it’s better to place safe and future-proof). Use `try` for `safe-exceptions`.

kirelagin:
One option might be to add `-L` option to follow symlinks, but, I’d say, that’s too much trouble. Instead, we can just add an example of an invocation that would follow symlinks (find, xargs, readlink, something like this...).

main/Main.hs
gromakovsky: Also apparently it's in a wrong commit :roll_eyes:

Lucus16:
I agree with using `safe-exceptions` and adding an example like `nixfmt $(readlink -f $(find -name '*.nix'))`. Don't have time for it right now though, feel free to fix them and ask for my review.

tomberek:
We plan to deprecate the directory mode directly in nixfmt and instead encourage use of treefmt.

infinisil:
See https://github.com/NixOS/nixfmt/issues/151

PR #59, opened 2020-02-15
Add REUSE header to BK piepline.yml

kirelagin:
To make reuse lint happy.


Lucus16:
Maybe reuse lint should be in CI?

kirelagin:
Would be good, yes.

yorickvP:
Superseded by #87

PR #60, opened 2020-02-16
Correct a typo in Utf8.hs

felixonmars:


Lucus16:
Thanks for catching that! This code will be replaced with a library soon though.

yorickvP:
Thanks! 3b2e1f4d2b15ce127f0b075acb954bd263cca05d

PR #66, opened 2020-03-11
Attempt to switch to haskell.nix, lts-15

yorickvP:


Lucus16:
This makes building GHC or adding a cache a requirement for building `nixfmt` within reasonable time. That is not acceptable, it increases the barrier of participation too much. I understand that using haskell from basic nixpkgs is a pain, but we'll have to deal with it.

Currently, the build fails because it can't fetch `nixfmt.cabal` from Hackage. It shouldn't be doing that in the first place. Rather than switching to another build system, I suggest we find out why it tries to download the file while it should be using the local one.

yorickvP:
@zhenyavinogradov the ghcjs stuff is from https://github.com/input-output-hk/hokey-pokey/tree/angerman/ghcjs-shell, shell.nix should just evaluate `nixfmt.components.all`. It's true that people building nixfmt are unlikely to have our cache and we don't have a public pin for the ghc builds

PR #68, opened 2020-04-29
[OPS-984] Deployment: run `aws cp` before `aws sync` to replace existing files

zhenyavinogradov:
`aws s3 sync` compares file sizes and timestamps to decide if a file in
the bucket needs to be replaced, but all files in nix store have
timestamp=1. So if a new file happens to have the same size as the old
one, `aws s3 sync` would decide that it does not need to be updated.
Copy files to the local directory updating their timestamps to the
current time to fix it

### DIFF
--- a/.buildkite/pipeline.yml
+++ b/.buildkite/pipeline.yml
@@ -21,6 +21,9 @@ steps:
     commands:
       - nix-build -A nixfmt-webdemo
       - nix run -f . awscli -c
-          aws s3 sync --delete result/ "$CDN_BUCKET"
+          aws s3 cp --recursive result/ "$CDN_BUCKET"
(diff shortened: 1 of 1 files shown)


zhenyavinogradov:
`--delete` is the key flag there, it deletes files that were present in the old revision, but don't exist in the new one, to make sure that bucket contents are exactly what we expect them to be

kirelagin:
Ok, then first run `cp`, then `sync --delete`, so that it will unconditionally replace all “good” files and then delete all “bad” ones. IIUC, the timestamps will always be the same, so they won’t play any role.

zhenyavinogradov:
Changed it to run `cp` before `sync` instead of resetting the timestamps

PR #75, opened 2020-11-29
Convert this project to use Nix flakes

notgne2:


zhenyavinogradov (changes requested):
There are errors in CI, pipeline steps need to be updated as well. Other than that, looks good

notgne2:
https://hydra.nixos.org/build/129701778 😠

Lucus16:
I'm okay with adding support for flakes but I want nixfmt to continue to be usable without using flakes so please don't remove or significantly change the existing Nix files.

SuperSandro2000:
> to continue to be usable without using flakes

This is what flake-compat does. You don't need to have flakes enabled for it.

flake.nix
```diff
@@ -0,0 +1,93 @@
+# © 2019 Serokell <hi@serokell.io>
+# © 2019 Lars Jellema <lars.jellema@gmail.com>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+{
+  description = "A Simple multi-profile Nix-flake deploy tool.";
```
0x4A6F, line 7: Looks like copypasta from deploy-rs, could use a better description:  A formatter for Nix code

PR #79, opened 2021-09-19
Correctly handle last lines with spaces

### LINKED ISSUE #78: nixfmt breaks trailing spaces in strings when doing "" to '''' conversion
nixfmt seems to make a1 like a2 and b1 like b2, while they are different strings in 
[testin.txt](https://github.com/serokell/nixfmt/files/7176993/testin.txt)

Lucus16:
Fixes #78

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -23,7 +23,7 @@ import Nixfmt.Types
   (Ann(..), Binder(..), Expression(..), File(..), Leaf, ParamAttr(..),
   Parameter(..), Selector(..), SimpleSelector(..), StringPart(..), Term(..),
   Token(..), TrailingComment(..), Trivia, Trivium(..), tokenText)
-import Nixfmt.Util (commonIndentation)
+import Nixfmt.Util (commonIndentation, isSpaces)
 
 prettyCommentLine :: Text -> Doc
 prettyCommentLine l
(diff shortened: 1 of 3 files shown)


Lucus16:
Hmm, this introduces some unnecessary simple strings:
```
-      description = ''
-        A macro to generate structures which behave like bitflags.
-      '';
+      description =
+        "A macro to generate structures which behave like bitflags.\n";
```

Lucus16:
Fixed now.

Lucus16:
@kirelagin Could you get this reviewed and merged?

src/Nixfmt/Pretty.hs
```diff
@@ -301,23 +301,32 @@ isIndented parts =
...
+-- cleanly represented in an indented string.
+lastLineIsSpaces :: [[StringPart]] -> Bool
+lastLineIsSpaces [] = False
+lastLineIsSpaces xs = case last xs of
+    [TextPart t] -> isSpaces t
+    _            -> False
+
+isInvisibleLine :: [StringPart] -> Bool
+isInvisibleLine []           = True
+isInvisibleLine [TextPart t] = Text.null $ Text.strip t
```
Heimdell, line 314: Why not `isSpaces`, like in the rest of the code?

src/Nixfmt/Util.hs
```diff
@@ -78,3 +79,6 @@ dropCommonIndentation unstrippedLines =
     in case commonIndentation (filter (/=empty) strippedLines) of
             Nothing          -> map (const empty) strippedLines
             Just indentation -> map (fromMaybe empty . stripPrefix indentation) strippedLines
+
+isSpaces :: Text -> Bool
+isSpaces = Text.all (==' ')
```
Heimdell, line 84: What if there are tabs? https://github.com/NixOS/nix/pull/2911

I suggest replacing it with `Text.all Data.Char.isSpace` (it even is already imported!).
```suggestion
isSpaces = Text.all isSpace
```

Heimdell (commented):
The whitespace detection lacks tabs, and some derivations use it. This might produce incorrect output.

src/Nixfmt/Util.hs
Lucus16: Nix handles lines consisting entirely of spaces different from lines that contain other kinds of whitespace. This function is intended to match that behavior but it could probably be better documented. I got stuck on finding good names for lines without any characters vs lines consisting of only but at least one space vs lines that could be either.

src/Nixfmt/Pretty.hs
Lucus16: This function is specifically to catch an edge case where strings are better represented as a simple `""` string, rather than an indented `''''` string. It is not about matching the behavior of Nix. I added this because nixfmt used to turn things like `"\n\n"` and `"\n  "` into indented strings.

Issue #81, opened 2021-11-27
Escape Characters


0x4A6F:
### Description

ofBorg failure with `undefined variable ‘PYTHON_EXECUTABLE’ at ./pkgs/development/compilers/nextpnr/default.nix:57:22` after `nixfmt` run.

### Small example input

```nix
  patchPhase = with builtins; ''
    # use PyPy for icestorm if enabled
    substituteInPlace ./ice40/CMakeLists.txt \
      --replace ''\'''${PYTHON_EXECUTABLE}' '${icestorm.pythonInterp}'
  '';
```

### Expected output

```nix
  patchPhase = with builtins; ''
    # use PyPy for icestorm if enabled
    substituteInPlace ./ice40/CMakeLists.txt \
      --replace ''\'''${PYTHON_EXECUTABLE}' '${icestorm.pythonInterp}'
  '';
```

### Actual output

```nix
  patchPhase = with builtins; ''
    # use PyPy for icestorm if enabled
    substituteInPlace ./ice40/CMakeLists.txt \
      --replace '''${PYTHON_EXECUTABLE}' '${icestorm.pythonInterp}'
  '';
```

Heimdell:
The bug is in parser. It ignores the `''` "escape" sequence and just splices `\n` out of `''\n` into the parsed string.

Lucus16:
The parser is correct:
```
*Nixfmt Nixfmt.Parser Text.Megaparsec> parse file "stdin" "'' ''\\'''${foo}' ''"
Right (File (Ann SOF Nothing []) (Term (String (Ann [[TextPart "'${foo}' "]] Nothing []))))
```
Note that the double backslash is for haskell to generate a single backslash.
The problem is likely that the pretty printer doesn't know about this edge case.

Lucus16:
Right now, reinserting escapes for characters that would otherwise compose to mean other escape sequences is handled by this fairly simple piece of code: https://github.com/serokell/nixfmt/blob/5e004dddb8e5cf6d3c974643bc0b1c3d696a4631/src/Nixfmt/Pretty.hs#L388-L391
That should probably be replaced with something that can handle any sequence of single quotes followed by up to two other characters. It may also need to be aware of whether it is followed by the end of the indented string or an interpolation. It's not an easy algorithm to figure out.

PR #82, opened 2021-12-01
Implement verify flag to check that the output parses and formats the same as the input

Lucus16:


Lucus16:
113 files in nixpkgs parse differently, but most of them because I'm not sufficiently normalizing strings yet.

Lucus16:
With normalized strings, the `--verify` flag finds only the file mentioned in #81 when run on nixpkgs.

yorickvP:
Please rebase :)

Issue #86, opened 2022-02-21
Misparsing 000.0


yorickvP:
### Description

The parser skips all leading zeroes, even when they are semantically meaningful. This leads to a semantics change.

### Small example input

```
(a: b: builtins.toJSON [a b]) 000.0
```

### Expected output

```
(a: b: builtins.toJSON [a b]) 0 0.0
```

### Actual output

```
(a: b: builtins.toJSON [ a b ]) 0.0
```

Issue #85, opened 2022-02-21
Parse error on .5e5


yorickvP:
### Description

The parser doesn't accept floats starting with `.`

### Small example input

```
.5e5
```

### Expected output

```
0.5e5
```

### Actual output

```
<stdin>:1:6:
  |
1 | .0e+0
  |      ^
unexpected end of input
expecting expression

```

yorickvP:
```
nix-repl> let e = "e"; in (a: b: c: builtins.toJSON [a b c]) .0e5e.55
"[0,\"e\",0.55]"
```

PR #109, opened 2022-10-25
floats: allow starting with ., prefer over int (nix compat)

### LINKED ISSUE #98: Support for floats without a leading zero
The file `default.nix` : 

```
{
  hello = .9;
}
```

gives:

```
$ nixfmt default.nix
default.nix:2:13:
  |
2 |   hello = .9;
  |             ^
unexpected ';'
expecting expression
```

An even simpler testcase is: `nixfmt <(echo '.1')`

`nixfmt` may want to prepend a leading zero in this case for consistency.

yorickvP:
Fixes #98, #86, #85

I'm very unhappy over having to copy this much code from megaparsec, but it doesn't export it. See https://hackage.haskell.org/package/megaparsec-9.2.2/docs/src/Text.Megaparsec.Char.Lexer.html#float

Not sure about the version bounds on scientific.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix nixfmt trying to allocate temp files that aren't used.
 * Nixfmt now accepts the '-' argument to read from stdin.
 * `nixfmt [dir]` now recursively formats nix files in that directory.
+* Float and int literal parsing now matches nix.
 
 ## 0.5.0 -- 2022-03-15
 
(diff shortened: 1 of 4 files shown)


Lucus16:
I'm thinking we should move this to its own module, `Nixfmt.Parser.Float`. I don't mind all the copying since it means the implementation is now open to be patched for any other inconsistencies we find.

Lucus16:
The version bounds seem fine, we can change them when needed. The `.0.0` is redundant though and the line is not in sorted position :p

PR #111, opened 2022-10-26
nixfmt cmd: support recursively formatting an entire directory (nix fmt compat)

### LINKED ISSUE #99: Support single '-' operand and take it as reading from standard input
Currently passing the argument `-` does not try to read the "file" from standard input.

E.g.:
```
nixfmt - < flake.nix
```

This is a common standard as suggested per [POSIX](https://pubs.opengroup.org/onlinepubs/009695399/basedefs/xbd_chap12.html#tag_12_02).

This is especially required if using nixfmt via flake formatter `nix fmt`. As `nix fmt` does not handle standard input by default, but you need to run `nix fmt -- -`. One could also argue that it would be neat if `nix fmt` supported this.

yorickvP:
Fixes #99 

https://github.com/NixOS/nix/blob/master/src/nix/fmt.cc#L42

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -96,6 +96,7 @@ checkFileTarget path = Target (readFileUtf8 path) path (const $ pure ())
 
 toTargets :: Nixfmt -> [Target]
 toTargets Nixfmt{ files = [] }    = [stdioTarget]
+toTargets Nixfmt{ files = ["-"] } = [stdioTarget]
 toTargets Nixfmt{ check = False, files = paths } = map fileTarget paths
 toTargets Nixfmt{ check = True, files = paths } = map checkFileTarget paths
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -4,13 +4,14 @@
  - SPDX-License-Identifier: MPL-2.0
  -}
 
-{-# LANGUAGE DeriveDataTypeable, NamedFieldPuns #-}
+{-# LANGUAGE DeriveDataTypeable, NamedFieldPuns, MultiWayIf #-}
 
 module Main where
 
 import Control.Concurrent (Chan, forkIO, newChan, readChan, writeChan)
 import Data.Either (lefts)
 import Data.Text (Text)
+import Data.List (isSuffixOf)
 import Data.Version (showVersion)
 import GHC.IO.Encoding (utf8)
 import Paths_nixfmt (version)
@@ -20,6 +21,7 @@ import System.Exit (ExitCode(..), exitFailure, exitSuccess)
 import System.IO (hPutStrLn, hSetEncoding, stderr)
 import System.Posix.Process (exitImmediately)
 import System.Posix.Signals (Handler(..), installHandler, keyboardSignal)
+import System.Directory (listDirectory, doesDirectoryExist)
 
 import qualified Data.Text.IO as TextIO (getContents, hPutStr, putStr)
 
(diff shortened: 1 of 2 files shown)


main/Main.hs
```diff
@@ -94,10 +110,11 @@ fileTarget path = Target (readFileUtf8 path) path atomicWriteFile
 checkFileTarget :: FilePath -> Target
 checkFileTarget path = Target (readFileUtf8 path) path (const $ pure ())
 
-toTargets :: Nixfmt -> [Target]
-toTargets Nixfmt{ files = [] }    = [stdioTarget]
-toTargets Nixfmt{ check = False, files = paths } = map fileTarget paths
-toTargets Nixfmt{ check = True, files = paths } = map checkFileTarget paths
+toTargets :: Nixfmt -> IO [Target]
+toTargets Nixfmt{ files = [] }    = pure [stdioTarget]
+toTargets Nixfmt{ files = ["-"] } = pure [stdioTarget]
```
Lucus16, line 115: Did you check if this case gets hit? I think there's a good chance the single dash will be interpreted as some kind of flag and not end op in the positional arguments.

main/Main.hs
```diff
@@ -64,6 +66,20 @@ data Target = Target
...
     }
 
+-- | Recursively collect nix files in a directory
+collectNixFiles :: FilePath -> IO [FilePath]
+collectNixFiles path = do
+  dir <- doesDirectoryExist path
+  if | dir -> do
+         files <- listDirectory path
+         concat <$> mapM collectNixFiles (((path <> "/") <>) <$> files)
+     | isSuffixOf ".nix" path -> pure [path]
```
Lucus16, line 76:
```suggestion
     | ".nix" `isSuffixOf` path -> pure [path]
```

main/Main.hs
```diff
@@ -64,6 +66,20 @@ data Target = Target
     , tDoWrite :: Text -> IO ()
     }
 
+-- | Recursively collect nix files in a directory
+collectNixFiles :: FilePath -> IO [FilePath]
+collectNixFiles path = do
+  dir <- doesDirectoryExist path
+  if | dir -> do
+         files <- listDirectory path
+         concat <$> mapM collectNixFiles (((path <> "/") <>) <$> files)
```
Lucus16, line 75: I think it's better to always use `</>` from `filepath`.
```suggestion
         concat <$> mapM collectNixFiles ((path </>) <$> files)
```

main/Main.hs
```diff
@@ -64,6 +66,20 @@ data Target = Target
...
+     | isSuffixOf ".nix" path -> pure [path]
+     | otherwise -> pure []
+
+-- | Recursively collect nix files in a list of directories
+collectAllNixFiles :: [FilePath] -> IO [FilePath]
+collectAllNixFiles paths = concat <$> (sequence $ collectNixFiles <$> paths)
```
Lucus16, line 81: I would also be okay with importing `concatMapM` from `extra`.
```suggestion
collectAllNixFiles paths = concat <$> mapM collectNixFiles paths
```

main/Main.hs
yorickvP: Yes.
Previous:
```nixfmt: -: openFile: does not exist (No such file or directory)```
New:
```
❯ dist-newstyle/build/x86_64-linux/ghc-8.10.7/nixfmt-0.5.0/x/nixfmt/build/nixfmt/nixfmt - <<< "hello"
hello
```

Lucus16 (approved):

PR #117, opened 2023-04-06
Attach leading comments to the following token

### LINKED ISSUE #32: Move comments to the syntax following them

Lucus16:
It looks like this approach will work to fix #32.


piegamesde:
What's the state of this? I've merged this into #118, looking pretty good so far. I think the idempotency issues are not related to the fix, but to the style changes made alongside — after adapting them to the RFC style they went away.

Lucus16:
The state is what it says. I need to get around to fixing the idempotence issues before this can be merged.

PR #124, opened 2023-05-16
Fix typo in README.md

wbehrens-on-gh:


yorickvP (approved):

yorickvP:
@rvem could you trigger CI for this PR?

rvem:
Buildkite doesn't allow running CI on PRs from forks easily. I guess we should switch to GitHub Actions

yorickvP:
We can maybe add some conditionals to the pipeline upload step so CI can run with a specific label

Issue #130, opened 2023-08-15
Too strict upper bounds


sternenseemann:
```
base >=4.12.0 && <4.17,
megaparsec >=9.0.1 && <9.3,
text >=1.2.3 && <1.3
```

text 2.0, base 4.17, megaparsec 9.3 work

SebTM:
```
Using Parsec parser
Configuring nixfmt-0.5.0...
CallStack (from HasCallStack):
  withMetadata, called at libraries/Cabal/Cabal/src/Distribution/Simple/Utils.hs:370:14 in Cabal-3.8.1.0:Distribution.Simple.Utils
Error: Setup: Encountered missing or private dependencies:
base >=4.12.0 && <4.17,
megaparsec >=9.0.1 && <9.3,
text >=1.2.3 && <1.3
```

Looks like still the same issue when trying to install using flake

PR #132, opened 2023-09-23
Release preparations

piegamesde:


piegamesde:
@Lucus16 how thorough do you want the change log to be? I had a look at the git log, and didn't find anything where I would say that this should be added to the change log.

Also `cabal gen-bounds` doesn't report anything. I'd like to fix #130, but I don't know how to find out what the highest working boundary is or how to test it.

Lucus16:
If you don't see anything missing, that's good, it means we kept it up to date when making changes. `cabal gen-bounds` apparently doesn't do what I thought it did. I guess you can manually bump the upper bounds to match the suggestions from that issue and rebuild to check they work.

MAINTENANCE.md
```diff
@@ -0,0 +1,15 @@
+<!-- © 2019 Serokell <hi@serokell.io>
+   - © 2019 Lars Jellema <lars.jellema@gmail.com>
```
Lucus16, line 2: Makes no sense for a file you wrote. Feel free to put your own name or something like "the nixfmt maintainer team" maybe but I'm not sure if that makes sense legally.

piegamesde: Yeah I don't really understand what this is about, so I just copied things over to make CI happy ^^

> I'm not sure if that makes sense legally

AFAIK none of this makes sense legally, IIRC copyright notices were a thing only prevalent in american-style jurisdictions, and I think even they lifted that requirement some time ago?

Lucus16: Because serokell didn't want to think too much about what is right and wrong for legal stuff like this we use a tool built by the FSFE to decide. I don't really like how it looks but I trust the FSFE to have come up with something that makes sense legally.
More info at https://reuse.software/faq/

piegamesde:
I updated the flake dependencies, let me know if I forgot to update something. (I don't really understand flakes tbh, so I might have missed something)

Lucus16:
I don't understand flakes either but it looks updates. :+1:

piegamesde:
AFAICT, the CI now fails building the webdemo due to compile errors in some dependency because of our version bumps. Not sure how to fix that though

yorickvP:
At some point it should be possible to move from ghcjs to the new ghc wasi backend. Maybe it's already possible?

yorickvP:
Update: I looked at the state of it and the [JS backend](https://mmhaskell.com/blog/2023/3/13/ghc-961-includes-javascript-backend) might be more promising

infinisil:
We're in the call and looked at this together with @piegamesde and @Sereja313

Issue #119, opened 2023-04-19
Have more CI checks


piegamesde:
As a contributor, I'd like to have CI actions which

- runs the project's tests suite
- checks for proper code formatting
- maybe also runs a simple linter over the code

PR #134, opened 2023-09-27
[OPS-1448] Replace buildkite with github actions

Sereja313:
Problem: The current buildkite-based ci is being replaced with github actions.

Solution: Add github ci config, remove buildkite config.


piegamesde:
Closes #119

rvem (approved):

rvem:
Don't forget to remove buildkite check for required and add `check`:)

.github/workflows/main.yml
```diff
@@ -0,0 +1,42 @@
...
+  check:
+    runs-on: [self-hosted, nix]
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: reuse lint
+        run: nix shell .#packages.x86_64-linux.reuse -c reuse lint
+
+      - name: build
+        run: nix build .#nixfmt-static
```
Lucus16, line 27: Probably should check that `.#nixfmt-webdemo` and `.#nixfmt-deriver` build as well though I'm not sure what the latter is. Maybe `.#awscli` too to be sure that the deploy action will work after nixpkgs upgrades. Put them all in separate checks if possible.

Lucus16 (changes requested):

.github/workflows/main.yml
Sereja313: Added checks for `nixfmt-webdemo` and `awscli`. `nixfmt-deriver` is a nix expression generated by `cabal2nix`, it was pushed into `cachix`, which I guess no one uses. I also added `hlint` and `stylish-haske`ll checks, but decided not to apply the changes to avoid large rebases in other pull requests.

Lucus16 (approved):

PR #141, opened 2023-11-10
Parse inherit names with quotes

### LINKED ISSUE #97: Quotes should be allowed after `inherit`

piegamesde:
Supersedes #107. Closes #97. Tested on top of #118 against the test suite there (but not the entire Nixpkgs).

Note that this is an *extremely quirky* corner of Nix syntax, at the mercy of the internals of the cppnix parser. I do not plan to be 100% compliant and exactly match the behavior. But all non-malicious edge cases should be covered now, and no invalid nix code should be parsed.


infinisil (approved):
Looks good to me!

PR #166, opened 2024-03-14
Focus on stable Nix with npins instead of Flakes

infinisil:
Focus on stable Nix and use [npins](https://github.com/andir/npins) for dependencies.
With this PR, the `flake.nix` relies on `default.nix`, rather than the other way around.
    
- Simplifies the whole thing due to not having to mess with `system`, see https://github.com/NixOS/nix/issues/3843
- Makes the lockfile _way_ smaller, see https://github.com/NixOS/nix/issues/7730
- flake-compat is not necessary anymore

I believe the only negative impact is that downstream Flake users can't use the `follows` mechanism anymore to override nixpkgs. Unfortunately there's no way to give a better error in such a case.

Instead what works is
```nix
import inputs.nixfmt {
  inherit system;
  nixpkgs = inputs.nixpkgs;
}
```

Though I don't really expect this to be very necessary.

As a side benefit, this allows passing in a patched Nixpkgs just using e.g. `pkgs.applyPatches`, without needing Nix to support such functionality like in https://github.com/NixOS/nix/pull/6530.


dasJ (approved):

Sereja313 (approved):

Mic92:
How can I override the nixpkgs version now? Before one could use:

```
{
   nixfmt.inputs.nixpkgs.follows = "nixpkgs";
}
```

Also ho can I do offline evaluation with npins? With flakes, I can `nix flake archive` to get a gcroot over all inputs.

Mic92:
Also is there a way to avoid the nixpkgs re-import? This can be costly in terms of memory? Is it maybe possible to just override nixfmt from nixpkgs with this version?

Mic92:
Much better:

```
with import <nixpkgs> {};
nixfmt.overrideAttrs (old: {
  name = "nixfmt-2024-03-15";
  src = fetchFromGitHub {
    owner = "NixOS";
    repo = "nixfmt";
    rev = "136005fc3dd34ff0d6af9f8c57cd1d3d7d342537";
    sha256 = "sha256-g0MTMQZeRlKFQiKdEVc1IQ73nNeRwu0+GtUIvcm2NV4=";
  };
})
```

Mic92:
Ah. There is actually: nixfmt-rfc-style

infinisil:
> Also how can I do offline evaluation with npins? With flakes, I can `nix flake archive` to get a gcroot over all inputs.

I didn't know about `nix flake archive`, but it's a hard sell, it's documentation is [pretty much just `FIXME`](https://github.com/NixOS/nix/blob/0fb5024d8df46a47f5367c5b0a51f0b2f6d50032/src/nix/flake-archive.md#description). After looking at the [source code](https://github.com/NixOS/nix/blob/0fb5024d8df46a47f5367c5b0a51f0b2f6d50032/src/nix/flake.cc#L1038-L1087) to understand what it does, it doesn't seem to gcroot anything. Also this won't work for offline builds anyways since it doesn't include any dependencies of the derivations, but let me know if I'm mistaken.

With stable Nix there's a relatively easy way to gcroot all inputs to a derivation using [`.inputDerivation`](https://github.com/NixOS/nixpkgs/blob/69fff9c6e9afb103110986c9331fdd7682dc5858/pkgs/stdenv/generic/make-derivation.nix#L582-L587) (introduced by me a while ago, admittedly also underdocumented):

```
nix-build -A inputDerivation -o inputs
```

Though this won't gcroot any Nix inputs, so if you do a gc you'll still need internet, so it's not a complete solution either. Would be nice to see something like npins that also makes this easy.

Mic92:
Well. It still has more documentation than `inputDerivation`, which doesn't appear in any manual.

Issue #143, opened 2023-12-01
reduce runtime closure size


tomberek:
Currently the runtime closure of the rfc101 branch includes ghc-9 references. This creates a 2GB+ runtime closure. It seems this is for the library outputs.

The below fixes this, but we may instead want to do multiple outputs or use another approach to remove these.

```
diff --git a/flake.nix b/flake.nix
index 83661b5..231cadc 100644
--- a/flake.nix
+++ b/flake.nix
@@ -25,6 +25,13 @@
           haskell = super.haskell // {
             packageOverrides = self: super: {
               nixfmt = self.callCabal2nix "nixfmt" src { };
+              nixfmt-bin = self.nixfmt.overrideAttrs (old: {
+                disallowedRequisites = [ self.ghc ];
+                postFixup = ''
+                  rm -rf $out/lib
+                  rm -rf $out/nix-support
+                  '';
+              });
             };
           };
         };
@@ -68,6 +75,7 @@
         packages = rec {
           default = nixfmt;
           nixfmt = pkgs.haskellPackages.nixfmt;
+          nixfmt-bin = pkgs.haskellPackages.nixfmt-bin;
           nixfmt-static = haskell.lib.justStaticExecutables nixfmt;
           nixfmt-deriver = nixfmt-static.cabal2nixDeriver;
           nixfmt-js = ghcjsPackages.callCabal2nix "nixfmt" src { };
```

SuperSandro2000:
This probably just requires a wrapping with justStaticExecutables

infinisil:
Indeed, and this is already the case for the `nixfmt-static` attribute. Could be made the default though.

Issue #161, opened 2024-03-13
Deprecate Hackage package


infinisil:
Nixfmt is currently [on Hackage](https://hackage.haskell.org/package/nixfmt), but I don't think that's necessary:

- While `pkgs.nixfmt` is [currently being populated by Hackage](https://github.com/NixOS/nixpkgs/blob/9b5ed8cb1ed4d0e4411b20ac0f67f33eaae01e97/pkgs/top-level/all-packages.nix#L40327), there's no reason this would be necessary to build a Haskell package in Nix. Indeed the new `pkgs.nixfmt-rfc-style` [works without Hackage](https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/ni/nixfmt-rfc-style/package.nix)
- Nixfmt doesn't really have an API surface that would be useful enough on its own. I consider it an end-user application. If some useful and reusable API surface emerges from the codebase, we can still consider packaging _that_ on Hackage.

So I'd propose deprecating the Hackage package and not updating it anymore. This way `nixfmt` on Hackage will always be the legacy Serokell version.

This will remove a bunch of maintenance burden from the formatter team.

Ping @Sereja313 @Lucus16 @yorickvP

Sereja313:
I agree

dasJ:
Somebody from the [maintainers team](https://hackage.haskell.org/package/nixfmt/maintainers/) probably has to do this

infinisil:
Yeah I could do that, should I do it already or should we wait until we have a first initial release? It might also be a good idea to update the Hackage page with a bit of information about the deprecation.

dasJ:
We should also move the `nixfmt-classic` package in nixpkgs first to make sure it doesn't point to a non-existing hackage package.

jfly:
We looked at this in today's meeting. We'd like to add a note to the existing hackage package (which will require releasing a 0.6.1). Then we'll mark it deprecated with no replacement.

infinisil:
This is now done: https://hackage.haskell.org/package/nixfmt

PR #167, opened 2024-03-14
Nix improvements

infinisil:
Various improvements to the Nix code:
- Use the new [`lib.fileset`](https://nixos.org/manual/nixpkgs/unstable/#sec-functions-library-fileset) library instead of `builtins.path` and regex based source filtering
- Fix https://github.com/NixOS/nixfmt/issues/143 by always using `justStaticExecutables` (doesn't mean the binaries are statically linked confusingly)
- Don't build the haddocks anymore, we won't need those anyways with https://github.com/NixOS/nixfmt/issues/161
- Remove `nixfmt-deriver`, doesn't seem necessary
- Switch to `haskellPackages.shellFor`, a bit cleaner
- Add [haskell-language-server](https://github.com/haskell/haskell-language-server) to the environment, makes development very nice!

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,8 +19,8 @@ jobs:
       - uses: cachix/install-nix-action@v26
 
       - name: reuse lint
-        run: nix shell .#packages.x86_64-linux.reuse -c reuse lint
+        run: nix-build -A packages.reuse && result/bin/reuse lint
 
       - name: build nixfmt
-        run: nix build -L .
+        run: nix-build
         if: success() || failure()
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -22,5 +22,5 @@ jobs:
         run: nix shell .#packages.x86_64-linux.reuse -c reuse lint
 
       - name: build nixfmt
-        run: nix build -L .#nixfmt-static
+        run: nix build -L .
```
dasJ, line 25: Is this really the correct command when your precious PR was about using stable nix? Iirc the nix command is not stable either

infinisil: Ah good point, I originally avoided this because of the `nix shell` above because there's no equivalent stable command, but really, it's just `nix-build .. && reuse lint`, changed it now :)

Sereja313 (approved):

tomberek (approved):
from meeting...

Issue #164, opened 2024-03-14
Simplify copyright and licensing


infinisil:
Currently CI enforces [REUSE 3.0](https://reuse.software/spec/), which enforces each file to have copyright and licensing information. This is a bit tedious and I'd rather avoid it if we can.

@Sereja313 also mentioned [on Matrix](https://matrix.to/#/!JHmxsrPoybcbBVDSjm:numtide.com/$Jguy8Ao_xkIq4VI8Ni0wjNJLi3i4gBft1KMQECeoujY?via=nixos.org&via=matrix.org&via=schreibt.jetzt):
> This is definitely not necessary.

Currently master has these copyright lines throughout the codebase:
- © 2019 Lars Jellema <lars.jellema@gmail.com>
- © 2019 Serokell <hi@serokell.io>
- © 2020 Serokell <hi@serokell.io>
- © 2022 Lars Jellema <lars.jellema@gmail.com>
- © 2022 Serokell <hi@serokell.io>
- © 2023 piegames <git@piegames.de>
- © 2024 piegames <git@piegames.de>
- © 2024 Silvan Mosberger <contact@infinisil.com>

All files use [MPL-2.0](https://spdx.org/licenses/preview/MPL-2.0.html), the Mozilla Public License 2.0.

I don't think there's a need to relicense, and the license is already inlined in [LICENSE](https://github.com/NixOS/nixfmt/blob/master/LICENSE).

So we just need to union the copyright lines and put that in e.g. the README:
- © 2022 Lars Jellema <lars.jellema@gmail.com>
- © 2022 Serokell <hi@serokell.io>
- © 2024 piegames <git@piegames.de>
- © 2024 Silvan Mosberger <contact@infinisil.com>

Sereja313:
I'm not sure about merging existing copyright notices. While there are no strict requirements for adding new copyright notices, MPL-2.0 appears to prohibit the removal or modification of existing ones:
```
3.4. Notices

You may not remove or alter the substance of any license notices
(including copyright notices, patent notices, disclaimers of warranty,
or limitations of liability) contained within the Source Code Form of
the Covered Software, except that You may alter any license notices to
the extent required to remedy known factual inaccuracies.

```

infinisil:
I guess it should be

- © 2019-2022 Lars Jellema [lars.jellema@gmail.com](mailto:lars.jellema@gmail.com)
- © 2019-2022 Serokell [hi@serokell.io](mailto:hi@serokell.io)
[...]

PR #171, opened 2024-03-21
Condense REUSE 3.0 copyright and license notes

infinisil:
As discussed in https://github.com/NixOS/nixfmt/issues/164, we want to avoid requiring copyright notices in each file and instead condense them into a single place.

All copyright holders agreed to doing this already, but it would still be good to get an approval on this PR.

Note that I didn't remove REUSE 3.0, but just kind of cheated by using `*` to set the copyright for all files (I hope that works, we'll see when CI runs..) :)

Ping @Sereja313 @gromakovsky @piegamesde @Lucus16

### DIFF
--- a/.reuse/dep5
+++ b/.reuse/dep5
@@ -1,34 +1,8 @@
 Format: https://www.debian.org/doc/packaging-manuals/copyright-format/1.0/
 
-Files: .github/ISSUE_TEMPLATE/formatting-suggestion.md
-Copyright: 2019 Serokell <hi@serokell.io>
-           2019 Lars Jellema <lars.jellema@gmail.com>
-License: MPL-2.0
-
-Files: npins/*
-Copyright: 2024 Silvan Mosberger <contact@infinisil.com>
-License: MPL-2.0
-
-Files: test/diff/*
-Copyright: 2024 piegames <git@piegames.de>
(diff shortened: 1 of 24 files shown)


tomberek (approved):

piegamesde (approved):

.reuse/dep5
```diff
@@ -1,34 +1,8 @@
...
-           2022 Lars Jellema <lars.jellema@gmail.com>
-License: MPL-2.0
-
-Files: flake.lock
-Copyright: 2020 Serokell <hi@serokell.io>
-           2022 Yorick van Pelt <yorickvanpelt@gmail.com>
+Files: *
+Copyright: 2019-2022 Serokell <hi@serokell.io>
+           2019-2022 Lars Jellema <lars.jellema@gmail.com>
+           2023-2024 piegames <git@piegames.de
```
Sereja313, line 6:
```suggestion
           2023-2024 piegames <git@piegames.de>
```
Looks like the closing bracket was lost during the copying

Sereja313 (approved):

Lucus16 (approved):

.reuse/dep5
infinisil: Ah nice catch, fixed!

Issue #165, opened 2024-03-14
Run test suite in CI


infinisil:
There's a [test suite](https://github.com/NixOS/nixfmt/tree/b265f30b1d99d124d66276f798eb8957c89c7a30/test) now since #118, but it's not getting run by CI yet.

PR #172, opened 2024-03-24
ci: run tests

tomberek:
https://github.com/tomberek/nixfmt/actions/runs/8412676563/job/23033814428#step:6:1334

Fixes: https://github.com/NixOS/nixfmt/issues/165

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -24,3 +24,6 @@ jobs:
       - name: build nixfmt
         run: nix-build
         if: success() || failure()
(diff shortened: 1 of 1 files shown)


dasJ (approved):

infinisil (approved):
We should integrate the tests into cabal in the future, see https://cabal.readthedocs.io/en/3.4/cabal-package.html#test-suites. This way the tests would get run for any `nix-build`

For now though, this is better than nothing, so :+1:

PR #177, opened 2024-04-01
Refactor flake.nix to remove flake-utils

toastal:
There is no reason to be sending downstream dependencies for a for loop. Includes missing devShells.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -23,5 +23,9 @@
       });
 
       checks = forAllSystems (system: result.${system}.checks);
+
+      devShells = forAllSystems (system: {
+        default = result.${system}.shell;
+      });
     };
 }


flake.nix
```diff
@@ -1,25 +1,31 @@
...
-    flake-utils.lib.eachDefaultSystem (
-      system:
-      let
-        result = import ./default.nix { inherit system; };
-      in
-      {
-        packages = result.packages // {
-          default = result;
-        };
+    { self, nixpkgs }:
```
infinisil, line 3: You removed `flake-utils`, but added `nixpkgs` (and that without an `inputs` entry, which then runs into https://github.com/NixOS/nix/issues/7422). And if you don't commit the lockfile as here, we'll run into further non-reproducibility. Alternatively if you do commit the lockfile, we'll have two Nixpkgs versions that need to be fetched for the flake.

So, I don't think we should do this.

flake.nix
```diff
@@ -1,25 +1,31 @@
...
+      });
+
+      checks = forAllSystems (system: result.${system}.checks);
 
-        checks = result.checks;
-      }
-    );
+      devShells = forAllSystems (system: {
+        default = result.${system}.shell;
+      });
```
infinisil, line 29: What's the motivation for this?

infinisil (changes requested):

flake.nix
infinisil: > How is the current version using the `builtins` dependencies of `flake-utils` being more reproducible than using the lib from Nixpkgs?

Just because in the current PR at least, Nixpkgs isn't pinned, while flake-utils is. Without committing the lockfile to pin it, people will get different results at different points in time.

> Seems most folks are used to passing in inputs.nixpkgs.follows thru their flakes

That doesn't actually work, the npins-pinned Nixpkgs can't be overridden like that (see https://github.com/NixOS/nixfmt/pull/166 for what works instead)

> might it just be simpler to manually write out `x86_64-linux`, et al. for `system.exposedFlakes` and remove all dependencies then? It’s more repetitious “code”, but has zero dependencies. I would be willing to write it out.

Actually one thing that `flake-utils` allows you is to use the https://github.com/nix-systems/nix-systems pattern, which allows consumers to work around the problem of Flakes requiring a hardcoded list of systems. This wouldn't be possible anymore if we hardcode the systems locally.

flake.nix
infinisil: Makes sense, sounds fine to have this :)

Personally I'm using https://github.com/nix-community/nix-direnv, which also caches the shell for speed

infinisil: Sounds good to me!

PR #178, opened 2024-04-01
mv .envrc → .envrc.example

toastal:
It isn’t a good idea to make a personal configuration file hard to work with since it’s being tracked by the repository. This will a) trigger ``direnv`` to fire off its allow script before a use gets a chance to see what arbitrary code is about to be executed on their behalf & b) make making modifications more difficult as it would be easy to accidentally commit those modifications *or* requires addings all sorts of unnecessary extra configuration to allow a safe place to put in custom environment variables.

The simple solution is to just leave an example for those that want it.


Janik-Haag:
We already had this discussion and the overall sentiment seems to be that the it's currently handle is fine. https://github.com/NixOS/nixfmt/pull/118#discussion_r1514230533

piegamesde:
Yes, we forgot to do that unfortunately.

infinisil:
I'd be okay with something like this though:

```bash
if [[ -f .envrc.local ]]; then
  source .envrc.local
else
  use_nix
fi
```

This will allow you to customise the behavior without changing the default.

dasJ:
You woldn't: https://github.com/direnv/direnv/issues/348#issuecomment-571168850

PR #181, opened 2024-04-02
flake: add existing-but-missing devShells.default

toastal:


dasJ (approved):

infinisil:
For future reference, this is split off from https://github.com/NixOS/nixfmt/pull/177

PR #184, opened 2024-04-04
Enforce hlint

infinisil:
Enforce [hlint](https://github.com/ndmitchell/hlint) in CI and apply all its suggestions. This just keeps the Haskell code a bit tidier


dasJ (commented):
The change itself looks reasonable enough, I cannot judge the Haskell code.

piegamesde (approved):

PR #185, opened 2024-04-04
Ensure formatted Nix and Haskell files

infinisil:
Ensures that all Nix files are formatted with `pkgs.nixfmt-rfc-style` (not the local version to avoid having to build the package for the dev environment!) and Haskell files are formatted with [fourmolu](https://fourmolu.github.io/) with some reasonable configuration.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -23,6 +23,9 @@ jobs:
       - name: hlint
         run: nix-build -A checks.hlint
 
+      - name: treefmt
+        run: nix-build -A checks.treefmt
+
       - name: build nixfmt
         run: nix-build
         if: success() || failure()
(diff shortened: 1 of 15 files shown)


piegamesde (approved):

infinisil:
We looked at this in the team meeting:

  - `let in` and imports look weird, let's switch to [fourmolu](https://fourmolu.github.io/), which can be customized
  - Customizations:
    - in-style: no-space
    - import-export-style: diff-friendly
  - @infinisil will update the PR afterwards
  - Point out that `treefmt` can be run in the readme (future: pre-commit hook)

infinisil:
@piegamesde Now using fourmolu, please merge if it looks good :)

piegamesde (approved):

piegamesde:
treefmt fails

PR #192, opened 2024-04-15
Fix duplicate Nixpkgs diff comment

infinisil:
A duplicate comment is posted when the action runs the first time, because the posted comment isn't reused.

This could be seen in https://github.com/NixOS/nixfmt/pull/185, and it shouldn't happen in this PR anymore.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -44,6 +44,7 @@ jobs:
 
       - name: Create or update comment
         uses: peter-evans/create-or-update-comment@v4
+        id: couc
         with:
           comment-id: ${{ steps.fc.outputs.comment-id }}
           issue-number: ${{ github.event.pull_request.number }}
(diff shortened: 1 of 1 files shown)


philiptaron (approved):
On Matrix, @infinisil provided this explanation of the code:

1. Every step in GitHub Actions can have outputs. There's one step which finds a comment with "Nixpkgs diff" in it (Find Comment), and its output is used for the next step to target (`comment-id: ${{ steps.fc.outputs.comment-id }}`)

2. `${{ ... }}` is GitHub specific syntax that [gets interpolated at runtime](https://docs.github.com/en/actions/learn-github-actions/contexts#about-contexts).

3. The `create-or-update-comment` action creates [an output named `"comment-id"`](https://github.com/peter-evans/create-or-update-comment?tab=readme-ov-file#outputs).

This PR makes them work together to have just one comment.

piegamesde (approved):

PR #193, opened 2024-04-16
Fix #188

### LINKED ISSUE #188: Reaching a stable list formatting requires formatting twice
### Description

Just a small thing I noticed while packaging stuff for Nixpkgs.

### Small example input

```
[ "hello"

]
```

### Expected output

```
[ "hello" ]
```

### Actual output
after first formatting
```
[
  "hello"
]
```
then after formatting again, it gives the correct
```
[ "hello" ]
```

piegamesde:
Fixes #188


infinisil:
For now we shouldn't open PRs from a fork because it doesn't seem to work with the Nixpkgs diff check (#194), please open another one from this repo directly.

infinisil:
Looking more closely into this, I don't think this is the right fix. See #201 instead

tomberek:
Closing. Using #201

PR #196, opened 2024-04-19
Nixpkgs diff for forks

infinisil:
This is super tricky stuff, but this should allow us to run the Nixpkgs diff for forks without safety issues (e.g. for #193). See https://securitylab.github.com/research/github-actions-preventing-pwn-requests/ for some background.

Also included is:
- https://github.com/numtide/treefmt/pull/282
- Some related Nix and CI improvements

Best reviewed commit-by-commit.

---

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -28,6 +28,11 @@ jobs:
 
       - uses: cachix/install-nix-action@v26
 
+      - uses: cachix/cachix-action@v14
+        with:
+          name: nixos-nixfmt
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
       - name: checks
         run: nix-build -A ci
 
@@ -67,6 +72,11 @@ jobs:
       - uses: actions/checkout@v4
 
       - uses: cachix/install-nix-action@v26
+
+      - uses: cachix/cachix-action@v14
+        with:
+          name: nixos-nixfmt
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
 
       - run: |
           ./scripts/sync-pr.sh \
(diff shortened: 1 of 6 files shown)


infinisil:
The downside is that we'll need separate PRs to test changes to CI, because the workflows are taken from the base branch of PRs, not the PR branch. This is also why CI won't run in this PR.

I did just create a PR to verify it works though: https://github.com/tweag/nixfmt/pull/3

piegamesde:
Could you make it that CI takes it from the current branch if it is not a fork?

infinisil:
I think that only works in a hacky way by having it trigger for both `pull_request` and `pull_request_target` events, which will then trigger two events for every PR and therefore doubling the number of workflows started, and then maybe we can cancel half of them somehow. Haven't tried this yet, but I kind of don't expect it to work nicely.

piegamesde:
I see. Could you maybe make this based on the branch name? I.e. only trigger that CI when the branch is called `ci` and then we'll use only that for CI changes?

infinisil:
I tried really hard, but it's just not reasonably possible. I managed to [kind of get it working](https://github.com/infinisil/github-test/blob/66ca6be737655772290659ba2a113ec67feba8f5/.github/workflows/blank.yml), but this is super hacky and broken in some cases.

So without that, I'm just testing it in a fork, and that does work: https://github.com/tweag/nixfmt/pull/3. And the `sync-pr.sh` script can also be run locally, works just fine.

tomberek (approved):
Looks straightforward. Ci build functions as expected.

infinisil:
Can confirm that it works, see https://github.com/NixOS/nixfmt/pull/193 :)

PR #201, opened 2024-05-14
Preserve more newlines

### LINKED ISSUE #188: Reaching a stable list formatting requires formatting twice

infinisil:
This fixes #188 in a better way, as an alternative to #193

Previously, newlines after the last list/attrs/let-in item would not be preserved. So e.g.

```nix
[
  "foo"

]
```

would wrongly turn into

```nix
[
  "foo"
]
```

Subsequently it would turn into

```nix
[ "foo" ]
```

This PR fixes it, such that the original is preserved in this case.

Whereas previously all the newlines were stripped already during parsing, in this new commit all newlines are preserved, which both fixes the above problem and simplifies the code.


piegamesde:
I am not really a fan of the diff and think that the format specification should be changed instead. The rule of not preserving trailing empty lines should be easy to formalize and IMO is a good heuristic for desired behavior.

infinisil:
@piegamesde The diff is a bit confusing, because all the extra lines it "adds" are ones that were already there before. E.g. the newline in [this section](https://github.com/infinixbot/nixpkgs/blob/be4d19ff1a9a327ae805fdb344470ed6450256fc/doc/doc-support/lib-function-locations.nix#L70-L75) was [previously removed](https://github.com/infinixbot/nixpkgs/blob/005c0eef7c353c1e4add522d1953faef2fbd17e8/doc/doc-support/lib-function-locations.nix#L79-L84), but is [now preserved](https://github.com/infinixbot/nixpkgs/blob/36f3ae68f655be60f26a065f6c65342d0f5bd040/doc/doc-support/lib-function-locations.nix#L79-L85).

So while this causes the result to have more lines, the diff is actually smaller!
- [Before](https://github.com/infinixbot/nixpkgs/compare/nixfmt-201~4..nixfmt-201~3) it's 32,453 changed files with 2,233,339 additions and 1,189,212 deletions.
- [After](https://github.com/infinixbot/nixpkgs/compare/nixfmt-201~4..nixfmt-201) it's 32,439 changed files with 2,233,635 additions and 1,184,971 deletions.

Reasons for having this:
- The code becomes much simpler
- It's consistent with the standard
- People can manually control these newlines, the formatter doesn't enforce adding them

Sereja313 (approved):

tomberek (approved):
Reviewed live during fmt meeting.

Issue #197, opened 2024-04-22
Formatting unquotes interpolation


mmarx:
Consider `foo.nix`:
```nix
''
  O=''\'''${A}'
''
```

Formatting changes the semantics of this, turning it into an interpolation:
```nix
''
  O='''${A}'
''
```

```
foo.nix: Parses differently after formatting. This is a bug in nixfmt. Please report it at https://github.com/NixOS/nixfmt

Before formatting:
Term (IndentedString (Ann [] [[TextPart "O='''${A}'"],[]] Nothing))

After formatting:
Term (IndentedString (Ann [] [[TextPart "O='''",Interpolation Term (Token (Ann [] (Identifier "A") Nothing)),TextPart "'"],[]] Nothing))
```

piegamesde:
This language is cursed beyond repair, and I see no hope in fixing this without a proper formal specification of its string syntax and semantics.

infinisil:
Bumping this higher, since it's a correctness bug and is kind of blocking for https://github.com/NixOS/nixpkgs/pull/322537. @dasJ looked into it in todays meeting, I'll copy the notes here too:

- We discovered the relevant bits from the Nix lexer [here](https://github.com/NixOS/nix/blob/052f1320ddf72d617e337479ff1bf22cb4ee682a/src/libexpr/lexer.l#L182-L205)
- These should be relatively directly portable to Haskell (current code around [here](https://github.com/NixOS/nixfmt/blob/c67a7b65906bd2432730929bd0e4957659c95b8e/src/Nixfmt/Parser.hs#L192-L207))
- @Sereja313 recommended [this tutorial on Megaparsec](https://markkarpov.com/tutorial/megaparsec.html) by Mark Karpov

dasJ:
More details from the debugging session: It's probably the parser. The example string can be minimized to `''\'''${A}`. Nixfmt parses this as:
```
Term
    ( IndentedString
        ( Ann []
            [
                [ TextPart "'''${A}" ]
            , []
            ] Nothing
        )
    )
```

while nix parses it as
```
"'\${A}\n"
```

It probably dies because the first 4 characters (`''\'`) get converted to `'` and the rest gets converted to `\${A}` (at least this is what nix does) while nixfmt converts the second part to `''${A}` which is incorrect. I have no idea why but this only happens when the two tokens are combined.

PR #208, opened 2024-06-26
Fix 197

dasJ:
Closes https://github.com/NixOS/nixfmt/issues/197


dasJ:
Diff looks like this is now in line with what [we currently have in nixpkgs](https://github.com/NixOS/nixpkgs/blob/master/pkgs/applications/editors/vim/plugins/overrides.nix#L517) and nix now properly parses the string rather than failing to parse the file.

infinisil (approved):
Oh that's simpler than I'd have expected, neat!

infinisil:
@dasJ Since you seem to have understood the code, can you explain why this fixes it in the commit message? Would be great to also briefly describe the problem in the commit message for future readers :)

piegamesde:
Better put in some code comments as well, they are more visible than the git log

dasJ:
I actually don't understand why it fixes that 🥲

src/Nixfmt/Parser.hs
```diff
@@ -196,6 +196,7 @@ indentedStringPart =
       ( chunk "''\\n"
           <|> chunk "''\\r"
           <|> chunk "''\\t"
+          <|> chunk "''\\'"
           <|> chunk "''\\"
           *> (Text.singleton <$> anySingle)
             <|> chunk "''$"
```
infinisil, line 202: Ohh, I'm now seeing the problem: The formatting here is very misleading, because it makes me believe that `<|>` binds stronger than `*>`, but in reality it's the other way around: `<|>` is [`infixl 3`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Applicative.html#v:-60--124--62-), while `*>` is [`infixl 4`](https://hackage.haskell.org/package/base-4.20.0.1/docs/Control-Applicative.html#v:-42--62-).

It makes more sense when we format it this way, also added a comment:
```haskell
chunk "''\\n"
<|> chunk "''\\r"
<|> chunk "''\\t"
<|> chunk "''\\'"
-- ''\ followed by any char throws away the ''\ part, leaving just the char
<|> chunk "''\\" *> (Text.singleton <$> anySingle)
<|> chunk "''$"
<|> chunk "'''"
<|> chunk "$$"
<|> try (chunk "$" <* notFollowedBy (char '{'))
<|> try (chunk "'" <* notFollowedBy (char '\''))
<|> someP (\t -> t /= '\'' && t /= '$' && t /= '\n')
```

infinisil:
You nerdsniped me into fully re-implementing it :sweat_smile:: https://github.com/NixOS/nixfmt/pull/210

Issue #77, opened 2021-03-10
Configurable indentation size


RealityAnomaly:
I prefer to indent my Nix expressions with four spaces rather than two as I find it more readable. It'd be nice to have this as a command-line option you can specify, with two as the default.

Quantenzitrone:
yes and a tabs option please
preferably even respecting the settings in [.editorconfig](https://editorconfig.org/)

infinisil:
While it has been configurable for a while using `-w`, since #118, it's a _soft_ limit, and not a hard one. We should also implement a way to configure the hard limit now.

quantenzitrone:
`-w` is for line length not indentation

dasJ:
Yep sorry, that was an oversight on my side :/

infinisil:
Ah I see, makes sense!

I guess for that we should use [editorconfig's `indent_size`](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size)

infinisil:
@omkumar312 Thanks! If you need any help, feel free to also join the [Nix Formatting Matrix room](https://matrix.to/#/#nix-formatting:nixos.org).

Note also that there's already an open draft PR for this: https://github.com/NixOS/nixfmt/pull/209

omkumar312:
Command-Line Option: Introduce a flag, such as --indent, that allows users to set their preferred indentation size. example of usage:  'nixfmt --indent 4'

Keep the default indentation at two spaces for compatibility but enable users to choose four spaces if they prefer.
This change will provide greater flexibility and accommodate different coding styles, improving overall readability.

WindFreaker:
Any plans to allow for tab indentation?

piegamesde:
Not before the language is fixed

infinisil:
The number of spaces for indentation has become configurable with https://github.com/NixOS/nixfmt/pull/209

PR #209, opened 2024-06-26
Make indentation configurable

toastal:
Makes it easier to override or, hopefully in the future, make configurable (weird you can configure the line length, but not indentation width).

Personally I find 2-space-indented code very difficult to read. I am maintaining a fork on Codeberg to help myself in my own code bases with this patch (but increased indentation width) but this means anyone that wants to also change the indentation only needs to patch a single line instead of multiple spots.  This could be fixed if Nix were to support tabs, but that isn’t the case. It also just seems better practice to define the indentation than `2` integers thru the code.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute. 

Edit by @infinisil:
I implemented configurable indentation into this PR and added tests for it.

### DIFF
--- a/src/Nixfmt/Predoc.hs
+++ b/src/Nixfmt/Predoc.hs
@@ -347,11 +347,11 @@ mergeSpacings Hardspace (Newlines x) = Newlines x
 mergeSpacings _ (Newlines x) = Newlines (x + 1)
 mergeSpacings _ y = y
 
-layout :: (Pretty a, LanguageElement a) => Int -> Bool -> a -> Text
-layout width strict =
+layout :: (Pretty a, LanguageElement a) => Int -> Int -> Bool -> a -> Text
+layout width indentWidth strict =
   (<> "\n")
     . Text.strip
-    . layoutGreedy width
+    . layoutGreedy width indentWidth
     . fixup
     . pretty
     -- In strict mode, set the line number of all tokens to zero
@@ -480,8 +480,8 @@ indent n = Text.replicate n " "
 type St = (Int, NonEmpty (Int, Int))
 
 -- tw   Target Width
-layoutGreedy :: Int -> Doc -> Text
-layoutGreedy tw doc = Text.concat $ evalState (go [Group RegularG doc] []) (0, singleton (0, 0))
+layoutGreedy :: Int -> Int -> Doc -> Text
+layoutGreedy tw iw doc = Text.concat $ evalState (go [Group RegularG doc] []) (0, singleton (0, 0))
   where
     -- Simple helpers around `put` with a tuple state
     putL = modify . first . const
@@ -496,7 +496,7 @@ layoutGreedy tw doc = Text.concat $ evalState (go [Group RegularG doc] []) (0, s
           case textNL `compare` nl of
             -- Push the textNL onto the stack, but only increase the actual indentation (`ci`)
             -- if this is the first one of a line. All subsequent nestings within the line effectively get "swallowed"
-            GT -> putR ((if cc == 0 then ci + 2 else ci, textNL) <| indents) >> go'
+            GT -> putR ((if cc == 0 then ci + iw else ci, textNL) <| indents) >> go'
             -- Need to go down one or more levels
             -- Just pop from the stack and recurse until the indent matches again
             LT -> putR (NonEmpty.fromList indents') >> putText textNL textOffset t
@@ -623,14 +623,14 @@ layoutGreedy tw doc = Text.concat $ evalState (go [Group RegularG doc] []) (0, s
                 _ -> grp
               (nl, off) = nextIndent grp'
 
-              indentWillIncrease = if fst (nextIndent rest) > lineNL then 2 else 0
+              indentWillIncrease = if fst (nextIndent rest) > lineNL then iw else 0
                 where
                   lastLineNL = snd $ NonEmpty.head ci
-                  lineNL = lastLineNL + (if nl > lastLineNL then 2 else 0)
+                  lineNL = lastLineNL + (if nl > lastLineNL then iw else 0)
           in fits indentWillIncrease (tw - firstLineWidth rest) grp'
               <&> \t -> runState (putText nl off t) (cc, ci)
         else
-          let indentWillIncrease = if fst (nextIndent rest) > lineNL then 2 else 0
+          let indentWillIncrease = if fst (nextIndent rest) > lineNL then iw else 0
                 where
                   lineNL = snd $ NonEmpty.head ci
           in fits (indentWillIncrease - cc) (tw - cc - firstLineWidth rest) grp
--- a/test/correct-indent-4.nix
+++ b/test/correct-indent-4.nix
@@ -1,0 +1,932 @@
+# A file correctly formatted with 4 spaces as indentation
+{ lib }:
+
+let
+    inherit (lib)
+        addErrorContext
+        assertMsg
+        attrNames
+        concatLists
+        concatMapStringsSep
+        concatStrings
+        concatStringsSep
+        const
+        elem
+        escape
+        filter
+        flatten
+        foldl
+        functionArgs # Note: not the builtin; considers `__functor` in attrsets.
+        gvariant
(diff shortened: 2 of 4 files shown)


dasJ:
Do you think you could also wire up a CLI option to fix #77? This way people could use whatever indentation they want without having to patch nixfmt

src/Nixfmt/Predoc.hs
```diff
@@ -337,7 +342,7 @@ mergeSpacings Break Softspace = Space
 mergeSpacings Break Hardspace = Space
 mergeSpacings Softbreak Hardspace = Softspace
 mergeSpacings (Newlines x) (Newlines y) = Newlines (x + y)
-mergeSpacings Emptyline (Newlines x) = Newlines (x + 2)
+mergeSpacings Emptyline (Newlines x) = Newlines (x + indentWidth)
```
piegamesde, line 345: Not everything that is a `+ 2`  also is an indentation :)

src/Nixfmt/Pretty.hs
```diff
@@ -612,7 +613,7 @@ instance Pretty Expression where
       absorbAbs _ expr | isAbsorbableExpr expr = hardspace <> group' Priority (absorbExpr False expr)
       -- Force the content onto a new line when it is not absorbable and there are more than two arguments
       absorbAbs depth x =
-        (if depth <= 2 then line else hardline) <> pretty x
+        (if depth <= indentWidth then line else hardline) <> pretty x
```
piegamesde, line 616: I think this one isn't an indentation width either

src/Nixfmt/Predoc.hs
```diff
@@ -534,7 +539,7 @@ layoutGreedy tw doc = Text.concat $ evalState (go [Group RegularG doc] []) (0, s
             -- [  # comment
             --   1
             -- ]
-            Text _ _ TrailingComment t | cc == 2 && fst (nextIndent xs) > lineNL -> putText' [" ", t]
+            Text _ _ TrailingComment t | cc == indentWidth && fst (nextIndent xs) > lineNL -> putText' [" ", t]
```
piegamesde, line 542: This line is a workaround for a very specific quirk. I don't remember all the details, but after reading its documentation comment above I am not entirely certain as to how to adapt this to other indentation lengths, or even if this is necessary for other indentations in the first place.

My gut feeling says that either the issue only happens at indentation width 2, or that the workaround only works for indentation width 2. In the latter case, changing the value might introduce idempotency bugs again.

src/Nixfmt/Pretty.hs
```diff
@@ -81,7 +82,7 @@ instance Pretty Trivium where
     comment (if isDoc then "/**" else "/*")
       <> hardline
       -- Indent the comment using offset instead of nest
-      <> offset 2 (hcat $ map prettyCommentLine c)
+      <> offset indentWidth (hcat $ map prettyCommentLine c)
```
piegamesde, line 85: I'm a bit unsure about the `offset` here. I know that it used to allow odd numbers in the past (due to the way comments were styled) prior to some refactorings, but currently it may the case that all offsets are just indentation levels. In that case the `offset` function should be redefined to only track the indentation *level*, and then multiply that with the indentation width only during rendering.

src/Nixfmt/Predoc.hs
```diff
@@ -130,6 +131,10 @@ instance (Pretty a, Pretty b) => Pretty (a, b) where
 instance (Pretty a, Pretty b, Pretty c) => Pretty (a, b, c) where
   pretty (a, b, c) = pretty a <> pretty b <> pretty c
 
+-- | Default indentation spacing.
+indentWidth :: Int
```
piegamesde, line 135: Generally the indentation width configuration –just like the line length limit– should be done in a way that it is only used by the rendering engine. That way, it may simply be passed from the CLI as a parameter to `layout`.

A global constant, as defined here, would always require to change the code and recompile to configure the indentation depth.

piegamesde (dismissed):
Indentation handling is one of the most brittle parts of Nixfmt currently. Any code that touches this should have a strong testing story to avoid regressions for non-default indentation widths

src/Nixfmt/Predoc.hs
infinisil: I understand this as a code refactoring for a first step towards making it configurable in the future. But now that I'm seeing that this is probably not entirely correct, it's probably best to make it configurable already, so that it can also be tested properly.

infinisil: And to make it configurable, we should probably use [editorconfig's `indent_size`](https://github.com/editorconfig/editorconfig/wiki/EditorConfig-Properties#indent_size)!

infinisil:
In the [last meeting](https://discourse.nixos.org/t/formatting-team-meeting-2024-08-06/50222) we decided to draft non-ready PRs. Feel free to undraft when it's ready

Perhaps you could also sync with @omkumar312 on this, who showed interest in this work in #77

piegamesde:
I don't think that this change meaningfully advances us towards the goal of having configurable indentation. The fundamental concerns in my review were not addressed. (Though apologies for not giving feedback earlier)

piegamesde:
I'm sorry but this really needs a sound and thorough testing story. *Especially* since there are suspected open edge cases which can only be found by fuzzing against lots of code like Nixpkgs. I recommend looking at https://github.com/NixOS/nixfmt/pull/224/commits/aff0520720cf913523d6e3a45dd7df8a1b2516fa, which not only introduces a CLI option but also does some refactoring which makes it significantly easier to pass values from the CLI through to the `layout` function.

infinisil:
> What is unready? I pushed my changes 2 months ago & never got another review

Oh didn't see that, sorry.

@piegamesde I don't think this needs testing as is, because it doesn't provide any user interface for it, or guarantee that it's gonna keep working. In essence, this PR just does an internal refactoring.

src/Nixfmt/Predoc.hs
```diff
@@ -2,6 +2,7 @@
 {-# LANGUAGE OverloadedStrings #-}
 
 module Nixfmt.Predoc (
+  indentWidth,
```
infinisil, line 5:
```suggestion
```

src/Nixfmt/Pretty.hs
```diff
@@ -21,6 +21,7 @@ import Nixfmt.Predoc (
   hardline,
   hardspace,
   hcat,
+  indentWidth,
```
infinisil, line 24:
```suggestion
```

infinisil (approved):
I do understand @piegamesde hesitance, but I think it's fine to merge this with the explicit non-guarantee of correctness or support for a different value. To treat this as purely an internal refactoring and effectively ignoring that you might patch over this.

But if you do patch over this and there's issues with it, I'll ask that you spend some effort to properly implement this feature, instead of disguising it as another internal refactoring, ok?

infinisil:
> Confirmation on?

This:

> But if you do patch over this and there's issues with it, I'll ask that you spend some effort to properly implement this feature, instead of disguising it as another internal refactoring, ok?

infinisil:
I'll look into the CI issue, not sure what's up with that

Vortriz:
Pinging, because there has been no update on this for the past 2 months.

infinisil:
Since the author doesn't seem to respond and I don't want this PR to linger around any longer, I implement the user-facing side of this myself now, so that this is not just an internal refactoring anymore and can be tested.

test/correct-indent-4.nix
```diff
@@ -0,0 +1,932 @@
...
+        split
+        toJSON
+        typeOf
+        ;
+
+in
+## -- HELPER FUNCTIONS & DEFAULTS --
+rec {
+    /**
+      Convert a value to a sensible default string representation.
```
piegamesde, line 60: The indentation here is wrong

PR #210, opened 2024-06-26
Re-implement indented string parsing to be more preservative and match the Nix one

### LINKED ISSUE #197: Formatting unquotes interpolation

infinisil:
This re-implements the indented string parser to fully preserve the input, while matching behavior to the Nix one. This fixes #197 as an alternative to https://github.com/NixOS/nixfmt/pull/208.

In addition to fixing that issue though, this also fixes cases like

```nix
''
  ''\a
  ''\
  ''\b
''
```

which was previously reformatted as

```nix
''
  a

    b
''
```


infinisil:
Whether something like `''\a` should actually be preserved or not could be debated, but for now I think it's best to just stick to the standard, which specifies that everything in a string should be preserved.

dasJ (approved):

PR #212, opened 2024-07-01
Predoc: Fix quadratic complexities in nest and offset

dasJ:
I honestly don't understand why this makes it faster, but for me the test suite runs ~2s faster now. This is an AI-generated contribution. The GPT explanation is:

By abstracting the transformation logic into a single function `transform`, we can pass different transformations for nesting and offsetting. This way, we avoid repeated traversals and ensure each element is processed exactly once.

This change ensures linear complexity for the `nest` and `offset` functions. We leverage the `map` function to traverse and transform the `Doc` structure only once, applying the necessary changes to each element.

### DIFF
--- a/src/Nixfmt/Predoc.hs
+++ b/src/Nixfmt/Predoc.hs
@@ -170,26 +170,32 @@ group x =
 group' :: (Pretty a) => GroupAnn -> a -> Doc
 group' ann = pure . Group ann . pretty
 
--- | @nest doc@ declarse @doc@ to have a higher nesting depth
+-- | @nest doc@ declares @doc@ to have a higher nesting depth
 -- than before. Not all nestings actually result in indentation changes,
 -- this will be calculated automatically later on. As a rule of thumb:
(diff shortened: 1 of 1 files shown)


piegamesde:
Someone who actually knows Haskell should vet this, but my gut feeling is that the complexity is still quadratic, and that the performance improvement is of other nature (or just measurement noise).

dasJ:
Now that I look at it, isn't `nest x` the same as `offset 1 x`?

piegamesde:
They're doing similar things but on different fields of `Text`

MangoIV:
I cannot verify this speedup locally and I would be very surprised, this changes anything significantly. Unlike chatgpt claims, this does not remove any quadratic complexity, either, it just called map transform now, if you resubstitute you can see that.

tchoutri:
From a cursory glance it appears that it is just a rephrasing of the previous code without real improvement on the data structures or operations.

infinisil:
Let's close this then

PR #216, opened 2024-07-09
Verify test

infinisil:
Test #215 (CI uses the workflow from the base branch, so we need another PR to test changes to CI)


infinisil:
Oof, looks like there's a problem, the Nixpkgs diff was still going after more than 2h30!

PR #217, opened 2024-07-11
Fix infinite recursion when walking Abstractions

infinisil:
This caused an infinite recursion when trying to minimise the non-idempotency when running --verify on

```nix
{
  foo
# bar
, baz # qux
}:
null
```

---

### DIFF
--- a/src/Nixfmt/Types.hs
+++ b/src/Nixfmt/Types.hs
@@ -345,7 +345,12 @@ instance LanguageElement Expression where
           )
     (Assert _ cond _ body) -> [cond, body]
     (If _ expr0 _ expr1 _ expr2) -> [expr0, expr1, expr2]
-    (Abstraction param _ body) -> [Abstraction param (ann TColon) (Term (Token (ann (Identifier "_")))), body]
+    -- If the body is just a token, it doesn't have an expression, only walk through the parameter
+    -- This is also the base case for the result below
+    (Abstraction param _ (Term (Token _))) -> walkSubprograms param
+    -- Otherwise, to separate the parameter from the body while keeping it a valid expression,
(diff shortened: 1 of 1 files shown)


src/Nixfmt/Types.hs
```diff
@@ -345,7 +345,7 @@ instance LanguageElement Expression where
           )
     (Assert _ cond _ body) -> [cond, body]
     (If _ expr0 _ expr1 _ expr2) -> [expr0, expr1, expr2]
-    (Abstraction param _ body) -> [Abstraction param (ann TColon) (Term (Token (ann (Identifier "_")))), body]
+    (Abstraction param _ body) -> walkSubprograms param ++ [body]
```
piegamesde, line 348: I'd prefer a fix that does the reduction as initially proposed, and merely fixes the missing recursion anchor. Roughly:

```suggestion
    (Abstraction param _ (Term (Token (ann (Identifier "_")))) -> walkSubprograms param
    (Abstraction param _ body) -> [Abstraction param (ann TColon) (Term (Token (ann (Identifier "_")))), body]
```

piegamesde (changes requested):
Please make the commit message more clear that this is about the `walkSubprogram`, because it is only used for debugging regressions and does not affect program execution otherwise

src/Nixfmt/Types.hs
infinisil: Ah I see, yeah that's better. I now changed it to:
https://github.com/NixOS/nixfmt/blob/4aaab5684d07ef8ea7ae75dd59fe005c034fb9cf/src/Nixfmt/Types.hs#L348-L353

piegamesde (approved):

PR #222, opened 2024-07-17
Always expand attrs

infinisil:
This implements this part of the standard: https://github.com/NixOS/nixfmt/blob/f09ce415c148aa56282252886cf421247a8bcc58/standard.md?plain=1#L1259-L1267

This inconsistency was noticed in https://github.com/NixOS/nixpkgs/pull/326430#discussion_r1679889039

Edit: Though there's also this here: https://github.com/NixOS/nixfmt/blob/f09ce415c148aa56282252886cf421247a8bcc58/standard.md?plain=1#L1149


infinisil:
As a better alternative to this, I think we should finally implement input-aware formatting of lists/sets. Specifically to never contract lists/sets that are expanded onto multiple lines already.

We previously added this to the standard to make that possible: https://github.com/NixOS/nixfmt/blob/f09ce415c148aa56282252886cf421247a8bcc58/standard.md?plain=1#L67

0x4A6F (approved):
Thanks

piegamesde:
The part of the standard you quote is taken out of context, this is a deliberate special case for bindings. IMO this sweeping approach of always expanding attributes is a net negative to the output layout, especially when attribute sets are non-last arguments to function calls.

Moreover, this misses several other similar inconsistencies which happen with lists.

The current rules have already been tweaked a lot, and I would strongly prefer an approach which refines them a bit more:

- Use `prettyTermWide` in `prettyApp` in contexts where `prettyTermWide` is already used (e.g. bindings)
- Expand the logic to also apply let binding bodies, in operator chains(?), and for lists

infinisil:
Closing in favor of the above alternative

PR #227, opened 2024-07-25
Support pipe operators

rhendric:
NixOS/nix#11131 landed :tada:

Let the debates about the proper formatting of these particular operators begin! Should they use the same rules as all other operators? This PR assumes yes, but there's an argument for sticking `<|` at the ends of lines rather than at the beginnings, since it is the final value in a chain of `<|`s that is special. :shrug:

### DIFF
--- a/src/Nixfmt/Types.hs
+++ b/src/Nixfmt/Types.hs
@@ -424,8 +424,8 @@ data Token
   | TLessEqual
   | TNot
   | TUnequal
-  | TPipeTo
-  | TPipeFrom
+  | TPipeForward
+  | TPipeBackward
   | SOF
   deriving (Eq, Show)
 
@@ -469,8 +469,8 @@ operators =
     [Op InfixL TAnd],
     [Op InfixL TOr],
     [Op InfixL TImplies],
-    [Op InfixL TPipeTo],
-    [Op InfixR TPipeFrom]
+    [Op InfixL TPipeForward],
+    [Op InfixR TPipeBackward]
   ]
 
 tokenText :: Token -> Text
@@ -523,6 +523,6 @@ tokenText TLess = "<"
 tokenText TLessEqual = "<="
 tokenText TNot = "!"
 tokenText TUnequal = "!="
-tokenText TPipeTo = "|>"
-tokenText TPipeFrom = "<|"
+tokenText TPipeForward = "|>"
+tokenText TPipeBackward = "<|"
 tokenText SOF = ""
--- a/src/Nixfmt/Types.hs
+++ b/src/Nixfmt/Types.hs
@@ -424,6 +424,8 @@ data Token
   | TLessEqual
   | TNot
   | TUnequal
+  | TPipeTo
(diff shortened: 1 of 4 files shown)


src/Nixfmt/Types.hs
```diff
@@ -424,6 +424,8 @@ data Token
   | TLessEqual
   | TNot
   | TUnequal
+  | TPipeTo
+  | TPipeFrom
```
piegamesde, line 428: Small nix: I'd prefer PipeRight and PipeLeft (or PipeForward/PipeBackwards) as names, as they're a bit more visual.

piegamesde (approved):
My intuition also says that putting the backpipes to the right might be beneficial, but until I've written any actual code to get a feeling for it I currently don't care. And this is undoubtedly the simpler implementation

test/diff/operation/out.nix
```diff
@@ -273,4 +273,17 @@
...
+    a // b
+    |> f "very long argument should justify splitting this over multiple lines"
+    |> g { }
+  )
+
+  (
+    g { }
+    <| f "very long argument should justify splitting this over multiple lines"
+    <| a // b
+  )
```
roberth, line 288: What I like about the `|>` formatting is that the functions, that have an equal role, are formatted equally.
This makes it easy to read. 

That is not the case here. This formatting makes it seem as if `g { }` is special and `f "..."` and `a // b` are similar, while the opposite is true.

```suggestion
  (
    g { } <|
    f "very long argument should justify splitting this over multiple lines" <|
      a // b
  )
```

A similar effect could be achieved by indenting the first function, but that's a little unusual, and it doesn't make the function operands look quite as equal.

```suggestion
  (
       g { }
    <| f "very long argument should justify splitting this over multiple lines"
    <| a // b
  )
```

infinisil: The first suggested style has the problem that lines might be arbitrarily long, which makes it unclear how the lines relate to, without looking to the right. This is why the standard [always puts operators in front](https://github.com/NixOS/nixfmt/blob/e819b2d0f9173f0c73d2e1fb4bc5155362046653/standard.md?plain=1#L765).

The second suggested style violates the [indentation rule](https://github.com/NixOS/nixfmt/blob/e819b2d0f9173f0c73d2e1fb4bc5155362046653/standard.md?plain=1#L232) of the standard.

But we also agree that it's not ideal for the first function to look different. For now we don't want to block on this and just go ahead with this PR as is.

It would also be possible for https://github.com/NixOS/rfcs/pull/148 to specify how it should be formatted (though it doesn't actually specify a `<|` operator right now).

infinisil (approved):
(reviewed during the formatting team meeting)

src/Nixfmt/Types.hs
infinisil: Done by @dasJ in https://github.com/NixOS/nixfmt/pull/227/commits/55feeb57b09c8f2df363e78e9842bf5d0c648936

PR #232, opened 2024-08-08
Allow building with newest GHC

nikolaik:
Hi there! Wanted to build nixfmt using the latest GHC (`brew install cabal-install`) and had to loosen the base version requirement to make it build


infinisil (approved):
Thanks!

PR #233, opened 2024-08-11
Don't expand lists within functions

piegamesde:
This is just a quick experiment to see where we can go with this.

### DIFF
--- a/test/diff/apply/out-pure.nix
+++ b/test/diff/apply/out-pure.nix
@@ -71,7 +71,7 @@
   }
   [
     (mapAttrsToStringsSep [
-      force
+      force # meow
       long
     ] "\n" mkSection attrsOfAttrs)
   ]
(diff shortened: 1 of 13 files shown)


test/diff/apply/out-pure.nix
```diff
@@ -69,12 +69,7 @@
     b = # Function call with comment
       mapAttrsToStringsSep "\n" mkSection attrsOfAttrs;
   }
-  [
-    (mapAttrsToStringsSep [
-      force
-      long
-    ] "\n" mkSection attrsOfAttrs)
-  ]
+  [ (mapAttrsToStringsSep [ force long ] "\n" mkSection attrsOfAttrs) ]
```
emilazy, line 72: nit: `force long` not very accurate any more

piegamesde: Yep, seen that. Need to switch that over to comments or something else which is guaranteed to be multiline across all possible formats to prevent this in the future.

piegamesde:
The format still looks a bit weird when the last two arguments are lists each. This can happen with partially applied string replacement, or with `mkRemovedOptionModule` for example. Not sure if it's worth special casing that though

infinisil (approved):

PR #234, opened 2024-08-16
[fix] remove space leaks in (,) by using strict state

MangoIV:
- benchmarks on maintainer-list.nix give a baseline of 88M max residency, nice triangle of death and a bit over 4 seconds duration
- heap profile shows a lot of memory allocated to (,)
- removing usagge of (,) in own code didn't affect the profile
- changing to strict state improves memory from 88 to 40 max residency and runtime goes down by 1 second

### DIFF
--- a/src/Nixfmt/Lexer.hs
+++ b/src/Nixfmt/Lexer.hs
@@ -5,7 +5,7 @@
 
 module Nixfmt.Lexer (lexeme, pushTrivia, takeTrivia, whole) where
 
-import Control.Monad.State (MonadState, evalStateT, get, modify, put)
+import Control.Monad.State.Strict (MonadState, evalStateT, get, modify, put)
 import Data.Char (isSpace)
 import Data.List (dropWhileEnd)
 import Data.Maybe (fromMaybe)
--- a/src/Nixfmt/Types.hs
+++ b/src/Nixfmt/Types.hs
@@ -44,7 +44,7 @@ module Nixfmt.Types (
   walkSubprograms,
 ) where
 
-import Control.Monad.State (StateT)
+import Control.Monad.State.Strict (StateT)
 import Data.Bifunctor (first)
 import Data.Foldable (toList)
 import Data.Function (on)


infinisil (approved):
Very nice, thanks a lot!

infinisil:
Can I ask how you're doing the benchmarking? I don't have a lot of experience doing that with Haskell

MangoIV:
> Can I ask how you're doing the benchmarking? I don't have a lot of experience doing that with Haskell

Heap Profiling: compile the program with profiling (e.g. by using the `--enable-profiling` with `cabal`), then run the program with heap profiling enabled in the RTS, e.g. to break down by cost centre and output to an eventlog, `+RTS -hc -l-au`, then use `haskellPackages.eventlog2html` to visualize the heap profile. 

Time profiling: almost the same, except you run with `+RTS -p -l-au` , then you can convert the eventlog to a json with `haskellPackaes.hs-speedscope` and load it into e.g. https://speedscope.app. 

The GHC manual also has a section about profiling haskell programs: 
https://downloads.haskell.org/ghc/latest/docs/users_guide/profiling.html

MangoIV:
lazy state is a red herring though, so that wasn't too hard, thankfully

infinisil:
Very useful, thanks!

MangoIV:
image

I have gotten `nixfmt` fast enough to feasibly run on hacke-packges.nix and even though the pretty printer is much faster than the parser, it  still got a pretty bad space leak

MangoIV:
The big brown triangle is labeled with `Int`, which means that it's probably some arithmetic that is not forced.

MangoIV:
image

and this is how it looks when it's gone.

MangoIV:
(yeah, the space leak of the parser is still pretty bad, I just now fixed `set` to force its arguments...)

MangoIV:
image

`DocE` uses lists, which is a really bad datatype for these sorts of things, so there's another space leak (orange triangle in the bottom right)

PR #236, opened 2024-08-17
Some more performance changes 

MangoIV:
- max residency is down by ~ 50%
- removed two space leaks that only show up on very large files
- speed is approximately the same 
- might be that it's even faster on really large files like hackage-packages.nix but I didn't run it often enough on to see 
- did some refactorings

### DIFF
--- a/src/Nixfmt.hs
+++ b/src/Nixfmt.hs
@@ -3,7 +3,6 @@
 module Nixfmt (
   errorBundlePretty,
   ParseErrorBundle,
-  Width,
   format,
   formatVerify,
   printAst,
@@ -22,9 +21,6 @@ import qualified Text.Megaparsec as Megaparsec (parse)
 import Text.Megaparsec.Error (errorBundlePretty)
 import Text.Pretty.Simple (pShow)
 
--- import Debug.Trace (traceShow, traceShowId)
-
-type Width = Int
 type Layouter = forall a. (Pretty a, LanguageElement a) => a -> Text
 
 -- | @format w filename source@ returns either a parsing error specifying a
@@ -36,7 +32,7 @@ format layout filename =
     . Megaparsec.parse Parser.file filename
 
 -- | Pretty print the internal AST for debugging
-printAst :: FilePath -> Text -> Either String Text
+printAst :: FilePath -> Text -> Either String a
 printAst path unformatted = do
   Whole unformattedParsed' _ <- first errorBundlePretty . Megaparsec.parse Parser.file path $ unformatted
   Left (unpack $ toStrict $ pShow unformattedParsed')
--- a/src/Nixfmt.hs
+++ b/src/Nixfmt.hs
@@ -1,5 +1,3 @@
-{-# LANGUAGE RankNTypes #-}
-
 module Nixfmt (
   errorBundlePretty,
   ParseErrorBundle,
@@ -28,8 +26,11 @@ type Layouter = forall a. (Pretty a, LanguageElement a) => a -> Text
 -- of @w@ columns where possible.
 format :: Layouter -> FilePath -> Text -> Either String Text
 format layout filename =
-  bimap errorBundlePretty layout
+  bimap errorBundlePretty f
     . Megaparsec.parse Parser.file filename
+  where
+    -- f !x = layout $ maybe () (error . show) (unsafeNoThunks x) `seq` x
+    f !x = layout x
 
 -- | Pretty print the internal AST for debugging
 printAst :: FilePath -> Text -> Either String a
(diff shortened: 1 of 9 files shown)


MangoIV:
this branch on maintainers.nix
image
origin/master on maintainers.nix
image

this branch on hackage-packages.nix (1 min 52 sec)

image


origin/master on hackage-packages.nix (5 min 17 sec)
image


no changes in test suite times afaict

MangoIV:
I'm actually still thinking about the `Seq` change, I don't know if it improved things memory wise, although I do think it's the more idiomatic datastructure here...

MangoIV:
image
seems like it's only `toFormatter` and `fixup` which are really bad, so perhaps fixable ...

MangoIV:
problem is that it probably fucks up `Seq`'s complexities... 

But then again, `List` will be as bad... 

Vector, too, and they don't even have pattern matching because it's such a bad idea... Maybe we should just rewrite the functions to something more idiomatic, at once...

MangoIV:
image
This is how it looks now...

src/Nixfmt/Predoc.hs
```diff
@@ -264,33 +265,36 @@ isComment _ = False
...
+      let unexpandSubtract t = unexpandSpacing' (subtract t <$> mn)
+      in case x of
+          txt@(Text _ _ _ t) -> (txt :<|) <$> unexpandSubtract (textWidth t) xs
+          (Spacing Hardspace) -> (Spacing Hardspace :<|) <$> unexpandSubtract 1 xs
+          (Spacing Space) -> (Spacing Hardspace :<|) <$> unexpandSubtract 1 xs
+          (Spacing Softspace) -> (Spacing Hardspace :<|) <$> unexpandSubtract 1 xs
+          (Spacing Break) -> unexpandSpacing' mn xs
+          (Spacing Softbreak) -> unexpandSpacing' mn xs
+          (Spacing _) -> Nothing
+          ((Group _ ws)) -> unexpandSpacing' mn $ ws <> xs
```
infinisil, line 297: This also changes the semantics a bit, because `unexpandSpacing' (Just -1) Seq.Empty` now returns `Just []` instead of `Nothing`. This is likely the cause behind the changes in https://github.com/infinixbot/nixpkgs/commit/2fb6f2c261a13bc1f72a7d2c8a338b57853bce95. Even if that would be better, let's have this PR only be internal refactorings.

src/Nixfmt/Predoc.hs
```diff
@@ -159,9 +158,11 @@ trailing t = [Text 0 0 Trailing t]
...
 group x =
   pure . Group RegularG $
-    if p /= [] && (isSoftSpacing (head p) || isSoftSpacing (last p))
-      then error $ "group should not start or end with whitespace, use `group'` if you are sure; " <> show p
-      else p
+    case pretty x of
+      ((hp :<| _) :|> lp)
+        | isSoftSpacing hp || isSoftSpacing lp ->
+            error $ "group should not start or end with whitespace, use `group'` if you are sure; " <> show p
+      _ -> p
```
infinisil, line 165: This also changes semantics a bit, since the first pattern only applies with at least 2 elements, whereas before it also applied with just one.

infinisil (changes requested):
Neat! This seems fine to me other than the semantic changes

src/Nixfmt.hs
```diff
@@ -28,8 +26,11 @@ type Layouter = forall a. (Pretty a, LanguageElement a) => a -> Text
 -- of @w@ columns where possible.
 format :: Layouter -> FilePath -> Text -> Either String Text
 format layout filename =
-  bimap errorBundlePretty layout
+  bimap errorBundlePretty f
     . Megaparsec.parse Parser.file filename
+  where
+    -- f !x = layout $ maybe () (error . show) (unsafeNoThunks x) `seq` x
+    f !x = layout x
```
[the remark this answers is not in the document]
MangoIV: I’m still a bit experimenting that’s why I converted back to draft. The function that’s commented out checks that some closure doesn’t contain thunks :)

src/Nixfmt/Predoc.hs
MangoIV: Sure, this was an accident, nice catch!

src/Nixfmt/Predoc.hs
MangoIV: Oh you’re right. It was sure a bit late 🙃

MangoIV:
I will close this and start over during the weekend, I will have a bit more time and be hopefully less tired :D 

Gotta remove this horrible triangle...

MangoIV:
mmh. it appears there's barely any huge wins like the lazy pair anymore for space improvements that also improve runtime...

PR #237, opened 2024-08-20
Don't expand already-contracted lists if they fit on a line

infinisil:
This fixes annoying previous expansions of e.g.

    imports = [
      (mkRenamedOptionModule [ "foo" "bar" ] [ "florp" "baz" ])
    ];

Simpler alternative to #233


infinisil:
From the meeting notes: Turns out that always allowing contracted lists also means that e.g. buildInputs don’t get expanded as desired (unless it’s already expanded) which makes the diff harder to read when adding elements in the future

PR #243, opened 2024-08-28
README: Add installation and integration docs

dasJ:


README.md
```diff
@@ -14,19 +14,119 @@ The established standard Nix formatting differs considerably from the original o
...
+  outputs = { nixpkgs, self }: {
+    formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt-rfc-style;
+  };
+}
+```
+
+### treefmt
+
+[treefmt](https://github.com/numtide/treefmt) can be used to format repositories consisting of different languages with one command.
+A possible configuration for `nixfmt` in `.treefmt.toml` looks like this:
```
infinisil, line 59:
```suggestion
A possible configuration for `nixfmt` in `treefmt.toml` looks like this:
```

README.md
```diff
@@ -14,19 +14,119 @@ The established standard Nix formatting differs considerably from the original o
...
-- To get the most recent version, install from master:
-
-      nix-env -f https://github.com/NixOS/nixfmt/archive/master.tar.gz -i
-
-- Nix with flakes
-
-      nix profile install github:NixOS/nixfmt
+`nixfmt` was used as the basis for the official Nix formatter with a standardized formatting.
+The new formatting differs considerably from the original one.
+This is why two packages are in nixpkgs: `nixfmt-classic` with the non-standardized formatting and `nixfmt-rfc-style`.
```
infinisil, line 23:
```suggestion
A recent nixfmt version is available as `pkgs.nixfmt-rfc-style` in Nixpkgs. The formatting of this version differs considerably from the original nixfmt that was used as the basis for the standardised official formatter, which is also still available as `pkgs.nixfmt-classic` for now, though unmaintained.
```

Just a rewording to make the distinction clearer

README.md
```diff
@@ -14,19 +14,119 @@ The established standard Nix formatting differs considerably from the original o
...
+
+{
+  environment.systemPackages = [ pkgs.nixfmt-rfc-style ];
+}
+```
+
+### From the repository
+
+It's also possible to install `nixfmt` directly from the repository using `nix-env`.
+This allows the newest changes to be used, which may not be equal to the `nixfmt` version used to format nixpkgs, so this should be done with care!
```
infinisil, line 38: This applies to all methods here. I think there should be an additional section referring to how the `nix-shell` in Nixpkgs provides the correct version for Nixpkgs. Alternatively, let's just drop this sentence, otherwise it implies the above would be the correct version. As an upstream project we shouldn't really have to document downstream uses.

```suggestion
```

README.md
```diff
@@ -14,19 +14,119 @@ The established standard Nix formatting differs considerably from the original o
...
+
+```nix
+# ...
+treefmt-nix.mkWrapper nixpkgs {
+  # ... other options ...
+  programs.nixfmt-rfc-style.enable = true;
+}
+```
+
+More information about configuration can be found in [the official README](https://github.com/numtide/treefmt-nix?tab=readme-ov-file#integration-into-nix).
```
infinisil, line 82:
```suggestion
More information about configuration can be found in [the README](https://github.com/numtide/treefmt-nix?tab=readme-ov-file#integration-into-nix).
```

Just a minor precaution, we don't want people to think that treefmt is also an official Nix project.

README.md
```diff
@@ -14,19 +14,119 @@ The established standard Nix formatting differs considerably from the original o
...
+### nix fmt
+
+[nix fmt](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-fmt) (which is a flakes-only feature) can be configured by adding the following to `flake.nix` (assuming a `nixpkgs` input exists):
+```
+{
+  outputs = { nixpkgs, self }: {
+    formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt-rfc-style;
+  };
+}
+```
```
infinisil, line 54:
~~~suggestion
```nix
{
  outputs =
    { nixpkgs, self }:
    {
      formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt-rfc-style;
    };
}
```
~~~

We should adhere to our own formatting :P

infinisil (approved):
Some minor improvements, but this is great!

Feel free to merge yourself when addressed :)

Issue #244, opened 2024-08-28
rather simple decimal floats are converted to scientific / "E" notation


WilliButz:
### Description

Literal floats that I'd consider short and simple enough to represent them in decimal notation are reformatted to scientific / e notation.
I didn't find any specific reasoning about this behavior and I find it quite unexpected, therefore this issue.

### Small example input

```nix
{ a = 0.2; b = 0.068; c = 0.009; }
```

### Expected output

```nix
{
  a = 0.2;
  b = 0.068;
  c = 0.009;
}
```

### Actual output

```nix
{
  a = 0.2;
  b = 6.8e-2;
  c = 9.0e-3;
}
```

infinisil:
Damn, I don't even think I knew that Nix supports scientific notation, nor that nixfmt rewrites floats in that way. Definitely should be fixed!

PR #246, opened 2024-08-29
Don't rewrite floats and don't break e+ syntax

infinisil:
Fixes https://github.com/NixOS/nixfmt/issues/244

- Parses floats into their text as-is, without rewriting anything. This means that floats like `0.068` aren't rewritten as 6.8e-2 anymore. However, `.0` is also not rewritten as `0.0` anymore.
- Fixes a parsing bug that caused `1.e+0` to be formatted as `1.e + 0`, which is invalid syntax

---


WilliButz (approved):

piegamesde (approved):

0x4A6F (approved):

adrian-gierakowski:
Would be great if we could land this in nixpkgs.

Issue #207, opened 2024-06-23
Formatting of selections in function arguments does not agree with RFC 166


oxalica:
### Description

Related: #153

I'm trying to apply RFC 166 style on [rust-overlay](https://github.com/oxalica/rust-overlay) and noticed this weird formatting.

### Small example input

```nix
{
  rust-channel-of-version = assertEq (rustChannelOf { channel = "1.48.0"; }).rustc stable."1.48.0".rustc;
}
```

### Expected output

```nix
{
  rust-channel-of-version = assertEq (rustChannelOf { channel = "1.48.0"; }).rustc
    stable."1.48.0".rustc;
}
```

### Actual output

```nix
{
  rust-channel-of-version = assertEq (rustChannelOf { channel = "1.48.0"; })
    .rustc stable."1.48.0".rustc;
}
```

It splits the selection in the first argument, but does not put the second arguments on a newline, which is really confusing to me. It looks like `(assertEq (..)).rustc` instead of `assertEq ((..).rustc)`

### Rationale

As described in [RFC 166](https://github.com/NixOS/rfcs/pull/166):

> ### Function application
> - In a function application chain, the first element is treated as the "function" and the remaining ones as "arguments".
> - As many arguments as possible must be fit onto the first line.
> - If there is at most one multi-line argument that can be absorbed and all other arguments before/after fit onto a single line respectively, then that multi-line argument is absorbed.

Thus the first argument should be inlined as a whole, and the second argument should be put on a new line (otherwise it would exceed the max width 100).

infinisil:
Looking into this. So far just a minimal reproducer with `nixfmt -w 8`:

Input:
```nix
a (b).c d."e"
```

Output:
```nix
a (b)
  .c d."e"
```

infinisil:
Trying to fix it with https://github.com/NixOS/nixfmt/pull/249

PR #249, opened 2024-08-29
Fix selections getting expanded too liberally

infinisil:
Try to fix https://github.com/NixOS/nixfmt/issues/207

---


infinisil:
This is a hacky fix, but it seems to work pretty well: Very small Nixpkgs diff with effectively just improvements, and it fixes the originally reported issue.

piegamesde (approved):

PR #240, opened 2024-08-26
Deprecate recursive mode

infinisil:
The recursive mode has caused problems because it doesn't do any filtering, which can mess with files in `.git` directories and elsewhere. While we could support sane implicit filters and an interface to filter explicitly, that adds complexity and maintenance burden.

Instead, we can promote the use of `treefmt` instead, a "formatting multiplexer", which supports file filtering by default. So `nixfmt` will only be the "backend" formatter, while `treefmt` is the frontend.

Previously discussed in a team meeting here:
https://github.com/NixOS/nixfmt/issues/151#issuecomment-2008026386

PR #252, opened 2024-09-11
feat: format the current directory by default

zimbatm:
The common case is to format folders, not stdin.

This changes the default to format the current directory, and introduces
a `--stdin` flag for users that want to wrap `nixfmt` into editors and
other tools.

See also https://github.com/NixOS/nix/pull/11438

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -47,6 +47,7 @@ data Nixfmt = Nixfmt
     width :: Width,
     check :: Bool,
     quiet :: Bool,
+    stdin :: Bool,
     strict :: Bool,
     verify :: Bool,
     ast :: Bool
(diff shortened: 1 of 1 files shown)


infinisil:
It looks like you're unaware of https://github.com/NixOS/nixfmt/pull/240, where we deprecated directory recursing in favor of treefmt taking care of that. I'm not yet 100% sure on the interactions with your PRs 🤔

piegamesde:
Please save your snark for elsewhere. Nixfmt of course still accepts passing multiple files via CLI arguments, as is common practice in such applications. It is up to the caller to do the right thing.

infinisil:
Well said @piegamesde, thanks

Also, last time I [touched treefmt](https://github.com/numtide/treefmt/pull/282), it only used a single process per 1024 files. It went through a rewrite by now, so I'm not sure if that's still the case, but if it's not that sounds like something to fix.

Kamillaova:
Okay, thanks
If treefmt used one process before, I think it works the same way now

PR #254, opened 2024-09-17
Don't expand final list arguments of functions

infinisil:
Applies #233 to last arguments too

Best review commit by commit

---

This work is sponsored by [Antithesis](https://antithesis.com/) :sparkles:

PR #257, opened 2024-10-03
Better list argument handling

piegamesde:
This kind-of-but-not really supersedes #254. It doesn't explicitly touch on list expansion itself, however the part 2 commit implicitly doesn't force-expand last list arguments anymore by virtue of not absorbing them anymore. List expansion will require a fine-tuning pass after this change, notably the question of whether to *never* expand the last argument if it's a list (as done in #254), because right now it will only do so if the second-to-last argument also is a list.


infinisil (approved):
Added an extra commit to fix the remaining cases of poor `mkRenamedOptionModule`, but with that I think this is looking good!

infinisil (approved):
Updated and approved together in today's meeting. This was the final blocker for a treewide Nixpkgs reformat

PR #264, opened 2024-11-21
Add `--filename` flag for overriding the filename when using stdioTarget

### LINKED ISSUE #263: Add a flag for overriding the filename of files passed through stdin
### Description

Tools that automatically format files (e.g. Neovim, Jujutsu with `jj fix`, etc.) often work by calling a formatter like Nixfmt by passing a file's contents to stdin and then getting the formatted result via stdout, which has the benefit that the caller would have complete control over the original file, instead of having the formatter potentially overwriting it.

However, a drawback is that when there is a formatting error, all context around the *filename* is lost, which could be frustrating when dozens of files are involved:
```console
$ jj fix
<stdin>:71:3:
   |
71 |   '';
   |   ^
unexpected '''
expecting expression
```

Some other formatters like `clang-format` allow the user to supply the actual filename via an CLI option (e.g. `--assume-filename`) while still accepting input via stdin, and I think it would be great if Nixfmt would have something similar.

pluiedev:
When `--filename path` is specified, any diagnostics will use the `path` instead of `<stdin>`.

Closes #263

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -12,6 +12,7 @@ import Data.ByteString.Char8 (unpack)
 import Data.Either (lefts)
 import Data.FileEmbed
 import Data.List (isSuffixOf)
+import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import qualified Data.Text.IO as TextIO (getContents, hPutStr, putStr)
 import Data.Version (showVersion)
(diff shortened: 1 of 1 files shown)


infinisil (approved):
LGTM, thanks!

Needs a `treefmt` to satisfy CI though :)

pluiedev:
Just did :D

tomberek (approved):

mweinelt:
Was super suprised to see `-f` and `--filename` not point to the file to format, but to something else entirely.

Wouldn't `--display-filename` and `-d` be the less surprising option here?

jfly:
> Wouldn't --display-filename and -d be the less surprising option here?

Good point. We'll be deprecating this option in https://github.com/NixOS/nixfmt/issues/305.

Issue #198, opened 2024-04-25
Avoid list absorbtion when appended by two more non-list-literal expressions


ShamrockLee:
### Description

"[Absorption](https://github.com/NixOS/rfcs/blob/master/rfcs/0166-nix-formatting.md#terms-and-definitions)" is an "an alternative layout" that reduces indentation level, according to NixOS/rfcs#166. That is, the specification allows such "alternative" alongside the indent-after-newline right-hand-side style.

Given a chain of evaluated-to-list expressions chained by [list concatenation operators (`++`)](https://nixos.org/manual/nix/unstable/language/operators.html) consisting of at least one of which is a multi-line literal list and two adjacent non-list-literal expressions, if the multi-line list literal gets absorbed, the adjacent non-list-literal terms will be squashed into the same line, producing a long, hard-to-read, and git-diff-unfriendly line of code.

Current implementation of `nixfmt` *sometimes* perform the above-mentioned absoption. In my opinion, such behavior should be avoided.

### Small example input

```nix
{
  buildInputs = [
    "ee"
    "fff"
    "gggggggg"
    "hhh"
    "iiiii"
    "jjjjj"
    "kkkkkkkkkkk"
  ]
  ++ lib.optional true "lllll-lllll"
  ++ lib.optional false "mmmmmmmm"
  ;
}
```

### Expected output

```nix
{
  buildInputs =
    [
      "ee"
      "fff"
      "gggggggg"
      "hhh"
      "iiiii"
      "jjjjj"
      "kkkkkkkkkkk"
    ]
    ++ lib.optional true "lllll-lllll"
    ++ lib.optional false "mmmmmmmm";
}
```

### Actual output

```nix
{
  buildInputs = [
    "ee" # To patch /bin/sh shebangs.
    "fff"
    "gggggggg"
    "hhh"
    "iiiii"
    "jjjjj"
    "kkkkkkkkkkk" # Required at build time by SingularityCE
  ] ++ lib.optional true "lllll-lllll" ++ lib.optional false "mmmmmmmm";
}
```

piegamesde:
Hotfix in #269. The rule for this special case is that anything goes on the RHS of the `++`/`//` as long as it fits onto the rest of the line. This probably should be reworked in a more structured and stricter way.

For now, I've simply excluded the RHS to be another operation in order to expand operation chains as desired. I think that for now this should be sufficient, I don't think there are many other scenarios where people might run into this.

Issue #116, opened 2023-04-05
Tab-indented multiline-strings are turned into single-line strings


infinisil:
Input (note that every line starts with a single tab character):

```nix
	''
	  foo
	  bar
	''
```

Expected output:
```nix
''
  foo
  bar
''
```

Actual output:
```nix
"  foo\n  bar\n"
```

This can be disastrous: https://gitlab.com/Fuzen-py/configsreborn/-/blob/2a6b5aa5c8bee37cc13ab35d186b8fa604f62b2c/Snow/homeManager/programs/waybar/default.nix#L93-96

Lucus16:
This does indeed generate ugly output. The thing is, your suggested expected output changes the contents of the string (it removes the leading spaces). In fact, this string is impossible to format as indented strings without those leading tabs. Those tabs are a bit of a hack and I don't want to generate code like that. The proper solution could be to use `concatStringsSep` or `concatMap` to add the spaces to a list of lines or it could be to remove the leading spaces, but that's a decision that the programmer should make. Nixfmt should never change the semantics of the code on its own. Generating function calls is also tricky because we can't be sure whether `lib` or `builtins` is bound to what we expect. In the end it's best to leave it to the user to detect these cases and manually choose a solution.

PS. If you lost access to the unformatted indented string and feel it's a lot of work to recover the indented string, you could remove the indentation of any single line in the string and reformat again and nixfmt will turn it back into an indented string since it now has a valid representation as indented string.

Lucus16:
I just checked `nix repl` and it seems Nix also includes the tabs as part of the string which means `nixfmt` is still changing semantics in this case. I'm considering making tabs in indentation a parse error.

dasJ:
This changed (probably from @infinisil's new multiline string handling). It now returns a 4-space indented multiline string:
```nix
''
    foo
    bar
''
```
or
```
00000000: 2727 0a20 2020 2066 6f6f 0a20 2020 2062  ''.    foo.    b
00000010: 6172 0a27 270a                           ar.''.
```

PR #269, opened 2024-12-02
Final polish

piegamesde:
Picking some low hanging fixes before the big treewide

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -603,7 +603,11 @@ absorbRHS expr = case expr of
   -- Special case `//` and `++` operations to be more compact in some cases
   -- Case 1: two arguments, LHS is absorbable term, RHS fits onto the last line
   (Operation (Term t) (LoneAnn op) b)
-    | isAbsorbable t && isUpdateOrConcat op ->
+    | isAbsorbable t
+        && isUpdateOrConcat op
+        -- Exclude further operations on the RHS
+        -- Hotfix for https://github.com/NixOS/nixfmt/issues/198
+        && case b of (Operation{}) -> False; _ -> True ->
         group' RegularG $ line <> group' Priority (prettyTermWide t) <> line <> pretty op <> hardspace <> pretty b
   -- Case 2a: LHS fits onto first line, RHS is an absorbable term
   (Operation l (LoneAnn op) (Term t))
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -62,6 +62,7 @@ import Nixfmt.Types (
   mapLastToken',
   tokenText,
  )
+import Nixfmt.Util (isSpaces)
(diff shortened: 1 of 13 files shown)


infinisil:
> before the big treewide

Oh um, but then we need to redo https://github.com/NixOS/nixpkgs/pull/359904, wait for another Hydra rebuild, then redo https://github.com/NixOS/nixpkgs/pull/361165

piegamesde:
Unfortunately, I stumbled upon #116 which I'd now consider a critical correctness issue and therefore blocking. I recently did a Nixpkgs treewide on fucked up strings (for a different variant of indented string syntax fuckup) and therefore know that such things actually still exist in the wild.

piegamesde:
As for "[Don't move trailing comments in let anymore](https://github.com/NixOS/nixfmt/pull/269/commits/124eca6d97ab3d0b52e57867b78faf15a9365c22)", my suggestion would be to keep the commit out and thus the old behavior for the initial formatting, and then add it back in. The other commits are strongly recommended to include for the treewide

infinisil (approved):
Checked the diff, looking good to me, thanks! The [third commit](https://github.com/infinixbot/nixpkgs/commit/4003d3a09fa5db03f749cb074db6803a861c9f38) some people might not like due to it introducing more lines, but I think it's the right move to fix the reported problem (https://github.com/NixOS/nixfmt/issues/198).

src/Nixfmt/Pretty.hs
```diff
@@ -536,6 +540,12 @@ isAbsorbable (Path _) = True
 -- Non-empty sets and lists
 isAbsorbable (Set _ _ (Items (_ : _)) _) = True
 isAbsorbable (List _ (Items (_ : _)) _) = True
+-- Empty sets and lists if they have a line break
+-- https://github.com/NixOS/nixfmt/issues/253
+isAbsorbable (Set _ (Ann{sourceLine = line1}) (Items []) (Ann{sourceLine = line2}))
+  | line1 /= line2 = True
+isAbsorbable (List (Ann{sourceLine = line1}) (Items []) (Ann{sourceLine = line2}))
+  | line1 /= line2 = True
```
Sereja313, line 548:
```suggestion
isAbsorbable (Set _ (Ann{sourceLine = line1}) (Items []) (Ann{sourceLine = line2})) = line1 /= line2
isAbsorbable (List (Ann{sourceLine = line1}) (Items []) (Ann{sourceLine = line2})) = line1 /= line2
```

I think this can be simplified a bit

Sereja313 (approved):

src/Nixfmt/Pretty.hs
piegamesde: Hm, while semantically equivalent, it has a slightly different intent and would break if someone added a slightly overlapping case later

Sereja313: Yeah it makes sense, feel free to ignore

PR #270, opened 2024-12-03
Don't move trailing comments in let anymore

### LINKED ISSUE #259: Suggestion: commented out lines at the end of a `let` block should stay in the `let` block
### Description

I often comment/uncomment swaths of code when developing. Since adopting `nixfmt`, I've found that sometimes those lines of code move around when I comment them, which makes uncommenting them a chore: I have to uncomment them, and then laboriously move them back to where they came from. I'd prefer if nixfmt just didn't move them around.

I haven't been able to find a past discussion about this. I have not read the entire RFC, but https://github.com/nix-rfc-101/rfcs/blob/master/rfcs/0166-nix-formatting.md#comments does say:

> - Anything after the first # of single-line comments must be preserved.
>   - This allows the common pattern of prefixing many lines with # to comment them out, without the formatter trying to change anything.

Which (to my eyes) looks like it's talking about this use case (commenting/uncommenting code).

### Small example input

(Yes this code is silly. It does reflect a simplified version of things that happen to me when I'm trashing about during development, though.)

```nix
let
  # unused = 42;
in
42
```

### Expected output

```nix
let
  unused = 42;
in
unused
```

### Actual output

```nix
let
in
# unused = 42;
42
```

Note how `unused = 42;` got moved out of the `let ... in` block. This is easier to see in gif form:

2024-10-20_09-31-31_pattern

piegamesde:
*Split out from #269 to delay it until after the Nixpkgs treewide*

-----

While this was initially introduced for good reasons, as can also be seen in the diff, that code was written with documentation comments in mind and not code comments. As per the bug report, the downsides were deemed to high to keep it in

Fixes #259


dasJ (approved):
Result looks good, cannot judge the code though

infinisil (approved):

PR #271, opened 2024-12-10
Enable `TypeFamilies` by default

poscat0x04:
For some reason nixfmt isn't building on my machine and GHC complains about a type equality constraint.


infinisil:
That's odd, the code base shouldn't depend on that extension, what error are you getting?

infinisil:
Can't really reproduce without more info, closing for now

PR #277, opened 2025-01-23
Add `scripts/nixfmt-mergetool`

oxij:
The script is self-documenting.

IMHO, this script should also be added to the output of `nixfmt-rfc-style` of Nixpkgs. Otherwise, after recent treewide changes, things became impossible to merge manually.

### DIFF
--- a/scripts/nixfmt-mergetool
+++ b/scripts/nixfmt-mergetool
@@ -1,0 +1,77 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 Jan Malakhovski <oxij@oxij.org>
+#
+# This file can be redistributed under the terms of Unlicense
+# <https://unlicense.org> license.
+
+usage() {
+    cat - << EOF
+usage: $0 LOCAL BASE REMOTE MERGED
+
+A \`git-mergetool\`-compatible merge tool that uses \`nixfmt\` to merge more
+stuff.
+
+This tool rejects all non-\`.nix\` files, for \`.nix\` files it simply
+
+- calls \`nixfmt\` on its first three inputs, followed by
+- running \`git merge-file\` on the same inputs, followed by
+- running \`nixfmt\` on the result.
+
+Put the following to your \`~/.gitconfig\`:
+
+\`\`\`
+[mergetool "nixfmt"]
+    cmd = nixfmt-mergetool "\$LOCAL" "\$BASE" "\$REMOTE" "\$MERGED"
+    trustExitCode = true
+\`\`\`
+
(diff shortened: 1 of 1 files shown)


scripts/nixfmt-mergetool
```diff
@@ -0,0 +1,77 @@
+#!/usr/bin/env bash
+# Copyright (c) 2025 Jan Malakhovski <oxij@oxij.org>
+#
+# This file can be redistributed under the terms of Unlicense
+# <https://unlicense.org> license.
```
infinisil, line 5: The only worry I have is the re-introduction of separately licensed files, which we moved away from in the past: https://github.com/NixOS/nixfmt/pull/171. Would you be okay with you to re-license it under the MPL-2.0 license and assign copyright to the generic "Nixfmt contributors" as defined in https://github.com/NixOS/nixfmt/blob/e825e956ae967ee24aa502a90099956d1f8bc1ed/.reuse/dep5#L3-L6?

```suggestion
```

infinisil (commented):
Neat, I didn't know about `git mergetool`! While this way of merging isn't fool-proof against syntactic errors, it is quicker than the [auto-rebase script](https://github.com/NixOS/nixpkgs/tree/master/maintainers/scripts/auto-rebase), and there's no problem with having both, so I say let's have this. Will need some work to integrate into the build, but we can do this in follow-up PRs.

For reference, here's an example of a syntactic error that can be introduced. Consider this file:

```nix
{
  foo

# , bar
}:
null
```

Formatting this on one branch turns it into

```nix
{
  foo,

# , bar
}:
null
```

But if on another branch the extra argument gets uncommented:
```nix
{
  foo

  , bar
}:
null
```

You'll end up with this invalid merge result (no merge conflict):
```nix
{
  foo,

  , bar
}:
null
```

This is an easily catchable syntactic error though. I can't imagine such a failure changing semantics, but would love to be proven wrong.

infinisil:
Because it's a bit tedious to distribute a shell script along with the build, I made a PR to implement this directly in the Haskell codebase: https://github.com/NixOS/nixfmt/pull/283. This also benefits from improved speed, because three less processes need to be spawned.

oxij:
On license: yes, feel free to re-license under whatever, I selected unlicense exactly for the reason it allows doing that.

On builtin `--mergetool`: nice!

infinisil:
Also just opened https://github.com/numtide/treefmt/issues/523, because this functionality would be useful for all formatters :)

infinisil:
Done in https://github.com/NixOS/nixfmt/pull/283

PR #291, opened 2025-04-02
Mergetool fixes

infinisil:
Turns out there are some problems to fix, see the commit messages for details. I made sure to add regression tests.

---

This work is funded by [Antithesis](https://antithesis.com/) and [Tweag](https://tweag.io) :sparkles:

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -33,15 +33,15 @@ import System.Console.CmdArgs (
   typ,
   (&=),
  )
-import System.Directory (doesDirectoryExist, listDirectory)
+import System.Directory (doesDirectoryExist, listDirectory, renameFile)
 import System.Exit (ExitCode (..), exitFailure, exitSuccess)
 import System.FilePath ((</>))
-import System.IO (IOMode (WriteMode), hPutStrLn, hSetEncoding, stderr, withFile)
+import System.IO (hPutStrLn, hSetEncoding, stderr)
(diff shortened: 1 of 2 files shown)


main/Main.hs
```diff
@@ -229,20 +221,20 @@ mergeToolJob opts@Nixfmt{files = [base, local, remote, merged]} = runExceptT $ d
               <$> formatTarget formatter (fileTarget path)
         )
 
-  (_, Just out, _, process) <- do
-    lift $
-      createProcess
-        (proc "git" ["merge-file", "--stdout", local, base, remote])
-          { std_out = CreatePipe
-          }
+  exitCode <- lift $ withFile merged WriteMode $ \out -> do
```
jfly, line 224: Feel free to ignore this comment:

My haskell is too weak to really grok this code. Is there any issue if we crash partway through invoking `git merge-file`? Could we end up with a cut off file?

I was expecting to see something like "write to a temp file and then atomically move". But I guess we didn't have that in the old code, or at least I don't see anything like that.

jfly (approved):
Thanks!

main/Main.hs
infinisil: Good point! I now added a commit that both simplifies the code and takes care of this. It works by not using the `--stdout` flag, causing `git merge-file` to write it to the `local` file directly (which is a temporary file created by `git mergetool`), and then move that file to `merged` after formatting :D

PR #293, opened 2025-04-13
Parse numbers as strings

### LINKED ISSUE #292: Out of bounds numbers get rewritten
### Description
Nixfmt rewrites out of bound number by truncating them to 64 bits.

```bash
~ ❯ echo '55555555555555555555555555555' | nixfmt                                                                                                                 
6430967150697920739
```
This turns most in nix invalid numbers into valid numbers, possibly masking errors.
The exeptions are `-9223372036854775808` and `9223372036854775808`, `I64_MIN` and `-I64_MIN` respectively. These get both turned into `-9223372036854775808` which nix cannot parse.(See https://github.com/NixOS/nix/issues/12899)

This also has the funny effect of adding double minus:
```bash
~ ❯ echo '-9223372036854775808' | nixfmt                                                                                                             
--9223372036854775808
~ ❯ echo '-9223372036854775809' | nixfmt                                                                                                             
--9223372036854775807
```

piegamesde:
We have no business in caring about what semantics Nix happens to ascribe to these numbers. For all we care about, they are opaque strings of digits.

Fixes #292


MattSturgeon (approved):
I can't review the haskell, but I approve of the concept.

infinisil (approved):

jfly (approved):

PR #295, opened 2025-04-29
update installation instructions

fricklerhandwerk:
this looks like a scary large change, but it really does the following:
- group installation methods by target (environment, editor, project)
- add Home Manager and `mkShell`
- simplify the examples for `treefmt-nix` and `git-hooks.nix`
- add admonitions for the samples that require `treefmt` in the
  environment
- remove `nix-env`, because that's the easiest way to discourage its use
- move `nix fmt` to the end, because it takes much space for little benefit
  (interactive use is impractical: invoking it requires a full evaluation every time,
  while all environment and auto-config options run instantly)

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,8 +281,6 @@ to return back to the unmerged state.
 
 ## Development
 
-You are encouraged to test this out on your code and submit any undesirable formatting you find as an issue.
-
 ### With Nix
 
 Haskell dependencies will be built by Nix.
--- a/README.md
+++ b/README.md
@@ -16,6 +16,10 @@ For more details, see the [RFC implementation tracking issue](https://github.com
 
 ## Installation
 
+> [!NOTE]
+> `nixfmt` can only process one file at a time.
+> Consider using a configuration helper for [formatting a project](#in-a-project).
+
 ### In the environment
 
 #### NixOS
--- a/README.md
+++ b/README.md
@@ -14,94 +14,155 @@ The established standard Nix formatting differs considerably from the original o
 
 For more details, see the [RFC implementation tracking issue](https://github.com/NixOS/nixfmt/issues/153).
 
-## Installation And Usage Instructions
+## Installation
 
-### nixpkgs/NixOS
+### In the environment
 
-`nixfmt` was used as the basis for the official Nix formatter with a standardized formatting.
-The new formatting differs considerably from the original one.
-A recent nixfmt version is available as `pkgs.nixfmt-rfc-style` in Nixpkgs.
-The formatting of this version differs considerably from the original nixfmt that was used as the basis for the standardised official formatter, which is also still available as `pkgs.nixfmt-classic` for now, though unmaintained.
+#### NixOS
 
-So installing this `nixfmt` is as simple as adding to the system packages:
+To install `nixfmt` on NixOS for all users, add it to the [`environment.systemPackages`](https://search.nixos.org/options?show=environment.systemPackages) configuration option:
 
 ```nix
 { pkgs, ... }:
-
 {
   environment.systemPackages = [ pkgs.nixfmt-rfc-style ];
 }
 ```
 
-### From the repository
+To install it on NixOS for a particular user, add it to the [`users.users.<user>.packages`](https://search.nixos.org/options?show=users.users.%3Cname%3E.packages) configuration option:
 
-It's also possible to install `nixfmt` directly from the repository using `nix-env`.
-Also, updates are not done automatically (as it would with the system packages).
-
-```console
-$ nix-env -i -f https://github.com/NixOS/nixfmt/archive/master.tar.gz
 ```
-
-### nix fmt
-
-[nix fmt](https://nix.dev/manual/nix/latest/command-ref/new-cli/nix3-fmt) (which is a flakes-only feature) can be configured by adding the following to `flake.nix` (assuming a `nixpkgs` input exists):
-```nix
+{ pkgs, ... }:
 {
-  outputs =
-    { nixpkgs, self }:
-    {
-      formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt-rfc-style;
-    };
+  users.users.example-user-name.packages = [ pkgs.nixfmt-rfc-style ];
 }
 ```
 
-### treefmt
+#### Home Manager
 
-[treefmt](https://github.com/numtide/treefmt) can be used to format repositories consisting of different languages with one command.
-A possible configuration for `nixfmt` in `treefmt.toml` looks like this:
+To install `nixfmt` in Home Manager, add it adding to the [`home.packages`](https://nix-community.github.io/home-manager/options.xhtml#opt-home.packages) configuration option:
+
+```nix
+{ pkgs, ... }:
+{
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -14,94 +14,155 @@ The established standard Nix formatting differs considerably from the original o
 
 For more details, see the [RFC implementation tracking issue](https://github.com/NixOS/nixfmt/issues/153).
 
-## Installation And Usage Instructions
+## Installation
```
MattSturgeon, line 17: This feels like a good place to mention that nixfmt itself doesn't support formatting whole directory trees, so it is usually recommended to use treefmt.

Perhaps reference the in-a-project section here too?

```suggestion
## Installation

> [!NOTE]
> nixfmt only supports formatting specific files; you cannot ask it to format a directory tree.
>
> If you want to be able to format your project efficiently, consider [configuring treefmt to run nixfmt](#in-a-project).
```

fricklerhandwerk: Yes, right.

```suggestion
## Installation

> [!NOTE]
> `nixfmt` processes one file at a time.
> Consider using a configuration helper for [formatting a project](#in-a-project).
```

With fewer words?

README.md
```diff
@@ -14,94 +14,155 @@ The established standard Nix formatting differs considerably from the original o
...
+    },
+})
+```
+
+> [!NOTE]
+> This only works when `nixfmt` is available [in the environment](#in-the-environment).
+
+### In a project
+
+#### `treefmt-nix`
```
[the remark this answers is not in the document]
fricklerhandwerk: Yes, probably, but I'd prefer to do it in a follow-up to keep the scope small. It already took quite a while to clean this up. Also we're linking to treefmt-nix docs here, so people can still find out the details even if we don't have them here.

README.md
MattSturgeon: How about:

> `nixfmt` can only process one file at a time.

?

README.md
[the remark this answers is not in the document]
MattSturgeon: > We're linking to treefmt-nix docs here, so people can still find out the details even if we don't have them here.

`nixfmt-tree` and `treefmt.withConfig` are unrelated to treefmt-nix, so there's no documentation regarding them over there.

> I'd prefer to do it in a follow-up to keep the scope small. It already took quite a while to clean this up. 

SGTM 👍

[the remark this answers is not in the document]
fricklerhandwerk: > `nixfmt-tree` and `treefmt.withConfig` are unrelated to `treefmt-nix`, so there's no documentation regarding them over there.

Ah okay I misread that. In those cases then, unless they are somehow better than (or at least en par with) `treefmt-nix` (UX-wise and maintenance-wise) I'd let them go the `nix-env` way.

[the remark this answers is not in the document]
MattSturgeon: The reason `nixfmt-tree` was introduced (recently) is because `nixfmt` was dropping support for formatting directories. Its target audience is users who want to use nixfmt without having to configure anything. The original intention was that nixfmt-tree would be promoted over using nixfmt directly.

The reason `treefmt-nix` exists is to serve as a "bank of configs" and a module system for configuring treefmt. It is a relatively large project, because it maintains modules for many formatting programs. The target audience is someone who doesn't mind managing additional dependencies, and wants to enable pre-defined configurations.

The reason `treefmt.withConfig` exists is to allow configuring `treefmt` with a rfc42 "structured freeform config", without needing external dependencies like `treefmt-nix`. It does not (currently) maintain a bank of configs for formatting programs. It comes built into nixpkgs and is lightweight and easy to bootstrap; no flake inputs, no `fetchtarball`, etc. The target audience is someone who likes configuring using nix and modules, but wants to keep things simple and minimise dependencies.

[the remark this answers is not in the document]
fricklerhandwerk: Got it, then `nixfmt-tree` should probably be at the top.

README.md
```diff
@@ -14,94 +14,159 @@ The established standard Nix formatting differs considerably from the original o
...
-`nixfmt` was used as the basis for the official Nix formatter with a standardized formatting.
-The new formatting differs considerably from the original one.
-A recent nixfmt version is available as `pkgs.nixfmt-rfc-style` in Nixpkgs.
-The formatting of this version differs considerably from the original nixfmt that was used as the basis for the standardised official formatter, which is also still available as `pkgs.nixfmt-classic` for now, though unmaintained.
+### In the environment
 
-So installing this `nixfmt` is as simple as adding to the system packages:
+#### NixOS
+
+To install `nixfmt` on NixOS for all users, add it to the [`environment.systemPackages`](https://search.nixos.org/options?show=environment.systemPackages) configuration option:
```
jfly, line 27: This is not your fault, but IMO this is doubly weird advice:

- We *just* told people `nixfmt` isn't that useful to use as it can only process one file at a time.
- It implies that people would want to format all nix code on their computer with the same version of `nixfmt`. This sort of bogus assumption doesn't apply when nix gives us good tooling to isolate dependencies to projects.

I would rather remove these two sections (NixOS and Home Manager) in favor of the "Declarative shell environment" example. If folks really want a system (or user) install of `nixfmt-rfc-style`, I'm confident they can figure that out once they know the package name.

README.md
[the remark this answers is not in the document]
jfly: We should add docs about `treefmt.withConfig`. I don't think it has to happen in this PR, though. I'll let @fricklerhandwerk decide: please either add it here, or file an issue for one of us formatting team members to do later.

Related tange:

> The target audience [of `treefmt.withConfig`] is someone who likes configuring using nix and modules, but wants to keep things simple and minimise dependencies.

I sure would love to not have to explain this to people. I'd prefer that we upstream treefmt-nix to nixpkgs and combine it with `treefmt.withConfig`. See proposal here: https://github.com/numtide/treefmt-nix/issues/349

README.md
```diff
@@ -246,8 +264,25 @@ git restore --merge FILE1 FILE2 FILE3
...
+    { nixpkgs, self }:
+    {
+      formatter.x86_64-linux = nixpkgs.legacyPackages.x86_64-linux.nixfmt-rfc-style;
+    };
+}
+```
+
 ## Development
 
+You are encouraged to test this out on your code and submit any undesirable formatting you find as an issue.
```
jfly, line 284: @mightyiam, FYI

piegamesde: I find that sentence a bit weird tbh. It feels hyper-specific and also a bit out of place, given that we also have CONTRIBUTING.md. Better to properly expand on our contribution flow in a separate PR instead of squeezing that in here

piegamesde: #300

README.md
jfly: @fricklerhandwerk and I discussed this, and we decided to leave this here. It *is* weird to install a specific version of nixfmt globally, but it's also necessary for the editor configuration we document below.

We could potentially add more words to explain this, but this is how things are before this PR. I'd rather merge this PR and iterate on this in the future.

jfly (approved):
LGTM. There's room for future improvement, which we'll deal with later.

README.md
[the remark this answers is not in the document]
jfly: I'm marking this resolved so we can merge this PR. Let's add `treefmt.withConfig` docs in the future.

PR #310, opened 2025-05-27
Add documentation about onboarding team members and swapping team leads

jfly:


team/README.md
```diff
@@ -38,3 +38,26 @@ For all non-trivial decisions:
...
   - Meeting notes
+
+## Onboarding
+
+### Team members
+
+- Add them to the [@NixOS/nix-formatting] GitHub team
+- Ensure they are on our Matrix channels:
+  - [#nix-formatting:nixos.org](https://matrix.to/#/#nix-formatting:nixos.org)
+  - [Internal discussion channel](https://matrix.to/#/!kEOZRmGPKpDpKcEBim:matrix.org?via=matrix.org&via=flausch.social)
```
MattSturgeon, line 49: Does this link need the `via` metadata?
```suggestion
  - [Internal discussion channel](https://matrix.to/#/!kEOZRmGPKpDpKcEBim:matrix.org)
```

MattSturgeon: Also: does this act as an invite, making the room effectively public? Or do the room settings prevent that?

MattSturgeon (approved):
Questions regarding the private room link, otherwise LGTM. Thanks!

team/README.md
jfly: Sounds like it still requires an invite.

I've removed the `via` stuff. This is just a reminder that the room exists, not a huge deal if the link is broken.

PR #311, opened 2025-05-31
Add a pre-commit hook for nixfmt with `nix run`

gauthsvenkat:
The pre-commit hook for `nixfmt` currently relies on `cabal` being installed on the user's system, which might not always be available.

This MR adds another hook (`nixfmt-nix`) that instead uses `nix run` to provide the formatting functionality.

Please let me know if this is not within this project's scope.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -8,7 +8,7 @@
 
 - id: &id-nix nixfmt-nix
   name: *id-nix
-  entry: nix --extra-experimental-features "flakes nix-command" run nixpkgs#nixfmt-rfc-style
+  entry: nix --extra-experimental-features "flakes nix-command" run nixpkgs#nixfmt
   language: system
   types:
     - nix


gauthsvenkat:
Hey @MattSturgeon,

> Is this just a typo or am I mis-understanding?

It is not a typo, you do need `cabal` installed for `pre-commit` to run `nixfmt`. For some languages (like Go or Python), `pre-commit` bootstraps the language (and/or the package manager) itself. Unfortunately, Haskell is not one of those languages.

> Wouldn't it be better to use the nixfmt-rfc-style packaged in nixpkgs?

That is an excellent idea. I was also wondering how to test if flakes are enabled, but it might be better to use `nix-shell`. 

> there's also the fact that we've since removed support for formatting directory trees.

I don't think that should be a problem, since `pre-commit` passes only changed files to the hook sequentially. Not sure if this is the default behavior, but I'm sure this mode can be configured. 

> If git-hooks will specify the filenames to format as CLI args, then it should be fine.

Yep, exactly. 

Let me update my branch and check if it works as intended with `nix-shell`.

gauthsvenkat:
Okay, there is an issue with using `nix-shell -p nixfmt-rfc-style'; It is not trivial to have `pre-commit` pass the file names to the executable specified by `--run` (or `--command`). 

The filenames are passed to `nix-shell` itself, and that doesn't work. Not sure how to work around this issue :/.

.pre-commit-hooks.yaml
```diff
@@ -4,4 +4,12 @@
   language: haskell
   types:
     - nix
-  description: The official (but not yet stable) formatter for Nix code
+  description: &desc The official (but not yet stable) formatter for Nix code
+
+- id: &id-nix nixfmt-nix
+  name: *id-nix
+  entry: nix run github:NixOS/nixfmt
```
jfly, line 11: This will track our `main` branch, which could cause issues for users. How about the version in nixpkgs?

```suggestion
  entry: nix --extra-experimental-features "flakes nix-command" run nixpkgs#nixfmt-rfc-style
```

(not tested, please verify this works!)

gauthsvenkat:
Not quite sure why CI is failing, I've resolved the merge conflicts.

jfly:
@gauthsvenkat, your PR was conflicting with https://github.com/NixOS/nixfmt/commit/b8c73c145a02d9011e891a0dce00d1e042ea46d7. I just rebased it and force pushed. Hopefully CI will pass (or at least fail in a different way :p).

PR #322, opened 2025-08-05
[#313] Send output of --ir to stderr

### LINKED ISSUE #313: --ir overwrites .nix files
From the help text:
```
     --ir             Pretty print the internal intermediate representation,
                      only for debugging
```
I assumed it would write the IR to stderr, but it overwrote all my nix files with the formatted AST instead. I had to run `grep -Po '\e\[0;94;1m[^\e]+\e\[0m' src/configuration.nix | sed -E 's/\x1b\[0;94;1m|\x1b\[0m//g' | nixfmt | sponge src/configuration.nix` to fix them all.

Sereja313:
Send --ir output to stderr, like --ast, to prevent overwriting source files with IR. Closes #313

### DIFF
--- a/src/Nixfmt.hs
+++ b/src/Nixfmt.hs
@@ -45,7 +45,8 @@ printAst path unformatted = do
 -- | Pretty print the internal IR for debugging
 printIR :: FilePath -> Text -> Either String Text
 printIR filename =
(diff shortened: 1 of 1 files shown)


jfly (approved):
LGTM!

MattSturgeon:
Maybe we should introduce a changelog file so that we remember to note fixes like this in the next release?

PR #324, opened 2025-08-05
Keep trailing comment in { and [

### LINKED ISSUE #323: Preserve more trailing comments
Today, we sometimes allow trailing comments, and sometimes move them to the next line. For example:

Input:

```
foo = { # this should be an SSID
}
```

Current output:

```
foo = {
  # this should be an SSID
}
```

Requested output:

```
# this should be an SSID
foo = {
}
```

We discussed this at today's team meeting, and agreed that we'd like to preserve more of these.

Originally reported below:

> Maybe there is a strong argument for assuming the comment should usually go above instead of below?

I'd argue:
- A comment at the end of the line is always *about that line*.
- Comments on separate lines should usually go *before* the thing that they concern.

There are probably edge-cases where just moving the comment to the next line is the right thing... but my guess is that the majority of cases is better handled by moving them to the previous line instead.

_Originally posted by @wolfgangwalther in https://github.com/NixOS/nixpkgs/issues/427460#issuecomment-3105894929_

piegamesde:
This is a partial revert of 5bb0639

The change introduced there is now redundant because we catch that specific corner case more precisely in the parser nowadays, so it's reasonable to revert

Fixes #323

CC @wolfgangwalther


test/diff/idioms_pkgs_4/out-pure.nix
```diff
@@ -163,8 +163,7 @@ in
       cc =
         let
           nativePrefix =
-            {
-              # switch
+            {  # switch
```
MattSturgeon, line 166: A lot of these test cases have two spaces


```suggestion
            { # switch
```

MattSturgeon (commented):
Concept SGTM

test/diff/idioms_pkgs_4/out-pure.nix
piegamesde: It is introduced deliberately here in order to prevent idempotency issues in some very rare special cases. The test suite may not make it look like it but the entirety of Nixpkgs hits that edge case only once. Take a look at the input here:

```
nativePrefix = { # switch
  i686-solaris = "/usr/gnu";
  x86_64-solaris = "/opt/local/gcc47";
}.${system} or "/usr";
```

Naively, it would be formatted like this:

```
nativePrefix =
  { # switch
    i686-solaris = "/usr/gnu";
    x86_64-solaris = "/opt/local/gcc47";
  }
  .${system} or "/usr";
```

Because the comment and `i686-solaris` are vertically aligned, the comment will be associated with the `i686-solaris` and not with the `{`. This kind of respecting vertical alignment is necessary because too many people write code like:

```
foo = # comment
          and now the code
```

and without respecting that vertical alignment the comment would not be associated with its correct code. Anyways, because of this, a second formatting would wrongly associate the `switch` comment with `solaris` (which was not the intent of the original code), and thus produce 

```
nativePrefix =
  {
    # switch
    i686-solaris = "/usr/gnu";
    x86_64-solaris = "/opt/local/gcc47";
  }
  .${system} or "/usr";
```

which is different from the output after one formatting, and thus an idempotency issue. The "fix" is to insert one single space before comments that should not be aligned with the line that follows, so that they will be parsed correctly.

MattSturgeon: Knowing it is a pre-existing behaviour makes me feel better. Although I'm unsure how I feel about the ugly hack now being more exposed by this PR.

I guess, although it is covered a lot in our test suite, in reality it should be fairly rare to run into comments that need to be explicitly mis-aligned with the following value?

jfly:
While we like the idea, somebody with time needs to figure out how to address the concerns in [this thread](https://github.com/NixOS/nixfmt/pull/324#discussion_r2255083128). Closing to reflect the fact that we're not actively working on this, but contributions are appreciated.

PR #325, opened 2025-08-05
experiment: never indent concatenation

piegamesde:


jfly:
Discussed in today's meeting with @piegamesde. While we're not opposed to this, none of us have the time/motivation to see this through, so we're going to close this to reflect the fact that we're not working on it. Contributions appreciated, though!

PR #341, opened 2025-09-19
fix: indent + operator when concatenating paths on new line

### LINKED ISSUE #327: Assignment to long attribute name (or attrset) of appended local path doesn't indent new line
### Description

When assigning an appended to long path with a very long left hand operand, the concatenating string is placed on a new line without an indentation.

### Small example input

```
let
  aLongVariableNameOrEvenDeepAttrsetGettingAssignedAnAppendedToLocalPath = ./. + "some/string/path/that/forces/a/wrap/to/happen";
in "example"
```

### Expected output

```
let
  aLongVariableNameOrEvenDeepAttrsetGettingAssignedAnAppendedToLocalPath = ./.
    + "some/string/path/that/forces/a/wrap/to/happen";
in
"example"
```

### Actual output

```
let
  aLongVariableNameOrEvenDeepAttrsetGettingAssignedAnAppendedToLocalPath = ./.
  + "some/string/path/that/forces/a/wrap/to/happen";
in
"example"
```

dyegoaurelio:
Closes #327

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -502,8 +502,12 @@ prettyApp indentFunction pre hasPost f a =
             (\fRendered -> group' RegularG $ fRendered <> line <> absorbLast a <> post)
       <> (if hasPost && not (null comment') then hardline else mempty)
 
-prettyOp :: Bool -> Expression -> Leaf -> Doc
-prettyOp forceFirstTermWide operation op =
+isPath :: Term -> Bool
+isPath (Path _) = True
+isPath _ = False
+
+prettyOp :: Bool -> Bool -> Expression -> Leaf -> Doc
+prettyOp forceFirstTermWide indentPlusOps operation op =
   let -- Walk the operation tree and put a list of things on the same level.
       -- We still need to keep the operators around because they might have comments attached to them.
       -- An operator is put together with its succeeding expression. Only the first operand has none.
@@ -526,7 +530,9 @@ prettyOp forceFirstTermWide operation op =
       prettyOperation (Nothing, expr) = pretty expr
       -- The others
       prettyOperation (Just op', expr) =
-        line <> pretty (moveTrailingCommentUp op') <> nest (absorbOperation expr)
+        if indentPlusOps && value op' == TPlus
+          then line <> nest (pretty (moveTrailingCommentUp op') <> absorbOperation expr)
+          else line <> pretty (moveTrailingCommentUp op') <> nest (absorbOperation expr)
   in group' RegularG $
       (concatMap prettyOperation . flatten Nothing) operation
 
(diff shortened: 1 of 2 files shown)


src/Nixfmt/Pretty.hs
```diff
@@ -526,7 +530,9 @@ prettyOp forceFirstTermWide operation op =
       prettyOperation (Nothing, expr) = pretty expr
       -- The others
       prettyOperation (Just op', expr) =
-        line <> pretty (moveTrailingCommentUp op') <> nest (absorbOperation expr)
+        if indentPlusOps && value op' == TPlus
+          then line <> nest (pretty (moveTrailingCommentUp op') <> absorbOperation expr)
+          else line <> pretty (moveTrailingCommentUp op') <> nest (absorbOperation expr)
```
jfly, line 535: I'm trying to understand the bug we're fixing: it seems to me that the old code *is* trying to "nest" (aka: indent?) the expression. Why didn't that actually happen?

jfly (approved):
I can't deny that this seems to fix the bug. Unfortunately, as I don't really understand the cause of the bug, I don't feel able to review the fix.

I'll leave this open for a few days to see if anyone provides insight, but after that, I'll err on the side of merging it.

jfly:
And thanks @dyegoaurelio for the contribution!

MattSturgeon:
IIUC, when we changed the formatting to avoid unnecessary indentation, the rule was that if the first element of a concatenation chain was a literal value on the same line as the binding `=`, then the concatenation lines wouldn't need indenting.

To me, the bug here is that the literal value (here, a path value) remains on the same line as the `=` binding, despite that line being beyond the soft line length limit. If it was pushed down to a new line, then the entire value would be indented.

src/Nixfmt/Pretty.hs
dyegoaurelio: The bug was that the operator itself wasn't being indented when it wrapped to a new line, only the expression that followed it

My change is moving the nest to wrap both the operator and the expression together, ensuring they're indented as a unit.

dyegoaurelio: I think so.

this is the only test case that changes if we remove `nest` from this function in

```diff
--- test/diff/idioms_lib_5/out.nix      2025-09-19 11:50:16.392324601 -0300
+++ /dev/fd/63  2025-09-20 15:55:35.886938291 -0300
@@ -347,8 +347,8 @@
         x:
         x == { }
         ||
-          # Accept {} for tests that are unsupported
-          (isDerivation x && x ? meta.timeout);
+        # Accept {} for tests that are unsupported
+        (isDerivation x && x ? meta.timeout);
       merge = lib.options.mergeOneOption;
     });
     timeout = int;
[ERROR] (run with --update-diff to update the diff)
```
(this is the change I made to master to get that behavior)
```diff
diff --git a/src/Nixfmt/Pretty.hs b/src/Nixfmt/Pretty.hs
index a383446..ed2f7d3 100644
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -526,7 +526,7 @@ prettyOp forceFirstTermWide operation op =
       prettyOperation (Nothing, expr) = pretty expr
       -- The others
       prettyOperation (Just op', expr) =
-        line <> pretty (moveTrailingCommentUp op') <> nest (absorbOperation expr)
+        line <> pretty (moveTrailingCommentUp op') <> absorbOperation expr
   in group' RegularG $
       (concatMap prettyOperation . flatten Nothing) operation
 

```

piegamesde:
Seeing your diff I had a hunch, and the quick test I did in #342 turned out to already be a proper fix for the issue. Basically, the code path that was causing the deindentation shouldn't have been hit in the first place, and by looking into its guard conditions it showed that `isAbsorbable` accidentally also included paths.

PR #348, opened 2025-10-23
Reformat indented strings to simple strings when appropriate

### LINKED ISSUE #274: Single-line multiline strings not containing " should be disallowed
### Description

Multiline strings that are a single line and do not contain the `"` character should be disallowed. This seems incorrect, we should insist on single-line strings where they apply.

### Small example input

```nix
''Hello world!''
```

### Expected output

```nix
"Hello world!"
```

### Actual output

```nix
''Hello world!''
```

dyegoaurelio:
Converts indented strings to simple strings when they contain no newlines, quotes, or backslashes. For example: ''hello'' becomes "hello" and ''path/to/file'' becomes "path/to/file".

Also handles escape sequence conversion (''$ -> \$ and ''' -> '') and updates standard.md to document this behavior.

closes #274

### DIFF
--- a/test/diff/string/out.nix
+++ b/test/diff/string/out.nix
@@ -13,11 +13,11 @@
    b
   "
   ###
-  ''''
+  ""
   ###
-  ''a''
+  "a"
   ###
-  ''${""}''
+  "${""}"
   ###
   ''
     ${""}
@@ -51,7 +51,7 @@
        e
   ''
   ###
-  ''''
+  ""
   ###
   ''
     declare -a makefiles=(./*.mak)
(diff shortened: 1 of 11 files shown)


test/diff/string/out.nix
```diff
@@ -92,4 +92,10 @@
   ''
 
   "--${"test"}"
+  "\${pkgs.ghostscript}/bin/ps2pdf"
+  "'test\$test"
+  "'test''quotes"
+  "'plain"
+  "between spaces "
+  "!#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~"
```
dyegoaurelio:
```
➜ nixfmt (reformat-idented-strings) ✗ nix repl
Nix 2.28.5
Type :? for help.
nix-repl> [                                                                                                   
            '' between spaces ''
            '' !#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~''
          ]
[
  "between spaces "
  "!#$%&'()*+,-./0123456789:;<=>?@ABCDEFGHIJKLMNOPQRSTUVWXYZ[]^_`abcdefghijklmnopqrstuvwxyz{|}~"
]


```

PR #349, opened 2025-10-23
Add nixfmt-static package

hatch01:
Added: `nix build .#nixfmt-static`

This adds the `nixfmt-static` package, which provides a statically linked `nixfmt` binary (built with musl for a smaller size).
It’s especially useful for tools like [[mason.nvim](https://github.com/mason-org/mason.nvim)], as it makes static binaries available for project releases across all distributions.

Let me know if I should include it in the GitHub pipelines.
Also, it would be great if we could add the binary to the current release!

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -56,11 +56,14 @@ let
     );
   };
 
-  build = lib.pipe pkgs.haskellPackages.nixfmt [
+  haskellBuildPipeline = [
     haskell.lib.justStaticExecutables
     haskell.lib.dontHaddock
     (drv: lib.lazyDerivation { derivation = drv; })
   ];
+
+  build = lib.pipe pkgs.haskellPackages.nixfmt haskellBuildPipeline;
+  buildStatic = lib.pipe pkgs.pkgsStatic.haskellPackages.nixfmt haskellBuildPipeline;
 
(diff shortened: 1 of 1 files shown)


default.nix
```diff
@@ -32,6 +32,16 @@ let
     config = { };
   };
 
+  pkgsStatic = import nixpkgs {
+    inherit system;
+    overlays = [ overlay ];
+    config = { };
+    crossSystem = {
```
MattSturgeon: Isn't `crossSystem` soft-deprecated in favour of specifying `buildPlatform` and `hostPlatform` separately?

Also: What's the advantage of instantiating a new nixpkgs instance vs using `pkgs.pkgsStatic`?

hatch01: Yeah, you are right, I simplified this!

hatch01:
Is it possible to add the binary built in the release please ?

infinisil:
We can't spend a lot of time on testing/maintaining this ourselves, but if you come up with something GHA-based that you can confirm works, I think we can accept that :)

Please note that our existing release process is documented in https://github.com/NixOS/nixfmt/blob/master/MAINTENANCE.md, so that needs to be adjusted too if you go for that

hatch01:
Currently @flashonfire is cooking a wonderful GHA for as automatic a release process as possible.

PR #350, opened 2025-10-23
some improvements wrt memory behaviour (and thus time behaviour, too) 

MangoIV:
I've tested this mainly on nixpkgs' hackage-packages.nix which is a quite large expression.
Previously: 
image

After removing some retainers: 
image

Runtime: 41s -> 13.7s (reduction by 2/3) 
Max residency: 750 Mb -> 380 Mb

As you can see, the largest space leaks (in the parser) are fixed, but the pretty printer still uses lazy lists rather recklessly so there's still a huge space leak (caused by list constructors) which I think, I will be able to fix, too, but I do not have more time tonight. 

This work was kindly made possible by well-typed.

### DIFF
--- a/nixfmt.cabal
+++ b/nixfmt.cabal
@@ -80,6 +80,7 @@ library
   hs-source-dirs:      src
   build-depends:
       base             >= 4.12.0
+    , containers
     , megaparsec       >= 9.0.1 && < 9.6
     , mtl
     , parser-combinators >= 1.0.3 && < 1.4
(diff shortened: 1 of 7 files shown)


MangoIV:
> I think it has something to do with the dependency you added


Yeah I assume I have to add that to the nix expr for this package but then also the formatter is failing (the second check)

MattSturgeon:
I've force-pushed a rebase with each commit formatted.

`nix-shell --run treefmt` was working fine for me, not sure what issue you were running into.

MangoIV:
Also note that the speedup means that the program was spending literally >2/3 of its time traversing the heap, it might be worthwhile to check if GC tuning improves time any more since it doesn’t really matter if the GC kicks in after the parsing (given that the space leak in the pretty printer is fixed) 

ideally the pretty printer should of course not allocate much anymore (except the output Text, but that’s negligible)

PR #353, opened 2025-10-30
Extend single-line list formatting beyond function arguments

### LINKED ISSUE #206: `--single-line-lists` option to format non-singleton lists on one line where possible
### Description

In the same spirit as `--width` (default 100) allowing end-users outside of nixpkgs to tweak nixfmt's behavior, I'd like an option to disable the default formatting of non-singleton lists.

When the option is enabled, lists of any element count should be rendered on a single line _unless_ that cannot be done for some other reason, such as comments within the list or the total line length being too long.

Alternatively, a `--max-list-items` option could be added:
- default 1 for the current behavior
- 0 would always produce a multi-line list
- a negative value (-1) would _always_ produce a single-line list (if possible, as described in the rest of this issue).
- Any other positive integer would use that value as the threshold for when to no longer attempt to format a list on a single line.

### Small example input

```nix
{
  single = ["a"];
  short = ["a" "b" "c"];
  long = ["a very very" "long list of" "a few elements" "with many many" "characters!"];
  shortWithComment = [
    # Elements
    "a" "b" "c"
  ];
  shortListWithAVeryVeryLongName_LoremipsumdolorsitametconsecteturadipiscingelitPhasellusaliquetutmauriseumollisSuspendisserhoncusdui = ["a" "b" "c"];
}
```

### Expected output

```nix
{
  single = [ "a" ];
  short = [ "a" "b" "c" ];
  long = [
    "a very very"
    "long list of"
    "a few elements"
    "with many many"
    "characters!"
  ];
  shortWithComment = [
    # Elements
    "a"
    "b"
    "c"
  ];
  shortListWithAVeryVeryLongName_LoremipsumdolorsitametconsecteturadipiscingelitPhasellusaliquetutmauriseumollisSuspendisserhoncusdui = [
    "a"
    "b"
    "c"
  ];
}
```

### Actual output

```nix
{
  single = [ "a" ];
  short = [
    "a"
    "b"
    "c"
  ];
  long = [
    "a very very"
    "long list of"
    "a few elements"
    "with many many"
    "characters!"
  ];
  shortWithComment = [
    # Elements
    "a"
    "b"
    "c"
  ];
  shortListWithAVeryVeryLongName_LoremipsumdolorsitametconsecteturadipiscingelitPhasellusaliquetutmauriseumollisSuspendisserhoncusdui =
    [ "a" "b" "c" ];
}
```

**Note**: Only the `short` list and the short list with a long name should be formatted differently when `--single-line-lists` is used.

dyegoaurelio:
Previously, single-line list formatting was limited to lists within function arguments. This change extends the behavior to all lists, applying the same rules:
- Lists with >6 items always: expand
- Lists with >1 items and at least one non-simple item: expand
- Otherwise: try to fit on one line

nixpkgs diff is empty because if the brackets are on different lines, we keep them like that (unless we're on`--strict` mode)

The diff on the test cases were positive IMO.

closes #206


test/diff/attr_set/out-pure.nix
```diff
@@ -370,10 +334,7 @@
       # Sans) than that horror.  But we do need the Adobe fonts for some
       # old non-fontconfig applications.  (Possibly this could be done
       # better using a fontconfig rule.)
-      [
-        pkgs.xorg.fontadobe100dpi
-        pkgs.xorg.fontadobe75dpi
-      ];
+      [ pkgs.xorg.fontadobe100dpi pkgs.xorg.fontadobe75dpi ];
```
MattSturgeon: This is a case where the expanded formatting is preferred:

```suggestion
      [
        pkgs.xorg.fontadobe100dpi
        pkgs.xorg.fontadobe75dpi
      ];
```

dyegoaurelio:
> Another heuristic may be that if the list is itself the value of an attribute binding, then it is likely to change over time. Although, this won't catch "foo = [ x y z ] ++ optionals bar [ a b c ]`, since in that example `[ a b c ]`is not being bound _directly_ to`foo`.

We're already using the `isSimple` check, which removes things like sets and functions with a lot arguments.

I think we could do something like this
```hs
allowedToContract (Token (LoneAnn (Identifier txt))) = txt  == "false" || txt == "true" -- allow only boolean literals.
allowedToContract (Selection {}) = False -- e.g pkgs.xorg
allowedToContract t = isSimple (Term t)
```


this prevents changes such as this one

```diff
index 063dc21..fd242fb 100644
--- a/test/diff/monsters_3/out-pure.nix
+++ b/test/diff/monsters_3/out-pure.nix
@@ -45,13 +45,7 @@ stdenv.mkDerivation rec {
     wrapGAppsHook4
     glib # for glib-compile-resources
   ];
-  buildInputs = [
-    cairo
-    glib
-    gtk4
-    libadwaita
-    pango
-  ];
+  buildInputs = [ cairo glib gtk4 libadwaita pango ];
```

while allowing changes such as this one
```diff
-  (map (
-    buildAllowCommand "allow" [
-      "snapshot"
-      "mount"
-      "destroy"
-    ]
-  ))
+  (map (buildAllowCommand "allow" [ "snapshot" "mount" "destroy" ]))
```

bentenjamin:
This is an awesome change, I really like the way it takes care of single brackets on their own line. I made this eldritch abomination (which is probably not very good nix code) that would benefit greatly:
```
  # Flatten this set/file into the format expected by firefox config file, for example `"browser.urlbar.trimHttps" = true"`
  # A functor allows a set to be called like a function. It's first argument is always itself.
  # Nesting the arguments like below allows this function to be called on either itself or another set
  # flatten = import ./settings.nix
  # settings = flatten {};
  # settings = flatten default-profile-settings-set;
  __functor =
    self:
    {
      settings ? self,
    }:
    let
      recurse =
        path:
        lib.concatMapAttrs (
          name: value:
          if builtins.isAttrs value then
            recurse (path ++ [ name ]) value
          else if builtins.isFunction value then
            { }
          else
            {
              ${
                builtins.concatStringsSep "." (
                  path
                  ++ [
                    name
                  ]
                )
              } =
                value;
            }
        );
    in
    recurse [ ] settings;
```

PR #354, opened 2025-11-01
Add automated CI workflow for releases

flashonfire:
Follow up PR after #349 

Added CI workflow that triggers when a version change is pushed (in nixfmt.cabal) to the master branch.
It automatically builds a static binary, tags the commit, and creates a GitHub release with notes pulled from CHANGELOG.md.

Example release: https://github.com/FlashOnFire/nixfmt/releases/tag/v4.1.0

Edit:
Just thought that, as an alternative, we could have a workflow that triggers when a release is published and simply appends the static binary to it. That way we can keep the release process as-is.
Let me know which option you prefer.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,102 @@
+name: Release on Version Change
+
+on:
+  push:
+    branches:
+      - master
+    paths:
+      - 'nixfmt.cabal'
+
+defaults:
+  run:
+    shell: bash
+
+jobs:
+  release-if-version-changed:
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout main
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 2  # need previous commit for comparison
+
+      - name: Get version from nixfmt.cabal
+        id: get_version
+        run: |
+          CABAL_FILE="nixfmt.cabal"
+
+          # Extract the version
+          VERSION=$(grep -m1 "^version:" $CABAL_FILE | awk '{print $2}')
+          echo "Version found in cabal: $VERSION"
+
+          # Get previous version from last commit
+          PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}')
+          echo "Previous version: $PREV_VERSION"
+
+          # Export for later steps
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "prev_version=$PREV_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Check if version changed
+        id: compare
+        env:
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+        id: get_version
+        run: |
+          CABAL_FILE="nixfmt.cabal"
+
+          # Extract the version
+          VERSION=$(grep -m1 "^version:" $CABAL_FILE | awk '{print $2}')
+          echo "Version found in cabal: $VERSION"
+
+          # Get previous version from last commit
+          PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}' || echo "none")
```
jfly: I'm worried about us failing to parse the file and then strange things happening. Could we remove the `|| ...` part?

```suggestion
          PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}')
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+          echo "Previous version: $PREV_VERSION"
+
+          # Export for later steps
+          echo "version=$VERSION" >> $GITHUB_OUTPUT
+          echo "prev_version=$PREV_VERSION" >> $GITHUB_OUTPUT
+
+      - name: Check if version changed
+        id: compare
+        run: |
+          if [ "${{ steps.get_version.outputs.version }}" != "${{ steps.get_version.outputs.prev_version }}" ]; then
```
jfly: It's important to pass inputs via env vars instead of directly interpolating into the shell script. Here's how:

```suggestion
        env:
          PREV_VERSION: ${{ steps.get_version.outputs.prev_version }}
          VERSION: ${{ steps.get_version.outputs.version }}
        run: |
          if [ "$PREV_VERSION" != "$VERSION" ]; then
```

Similar other interpolations below.

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+        if: env.version_changed == 'true'
+        uses: cachix/cachix-action@v14
+        with:
+          name: nixos-nixfmt
+          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+
+      - name: Build static binary
+        if: env.version_changed == 'true'
+        run: |
+          nix build -L .#nixfmt-static
```
jfly: Let's use stable nix here instead of flakes:

```
```
```suggestion
          nix-build -A packages.nixfmt-static
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,111 @@
...
+          mkdir -p release
```
jfly: Do we use this `release` directory? Can we remove this line?

Issue #305, opened 2025-05-13
Implement treefmt's (nascent) Stdin Spec


jfly:
treefmt will be adding a Stdin Spec. See https://github.com/numtide/treefmt/issues/573#issuecomment-2877869301 for details.

We should implement it, which means deprecating and removing the current behavior:

1. Read from stdin if no positional arguments are passed.
2. Use the filename specified via `--filename`: https://github.com/NixOS/nixfmt/pull/264

The new behavior will (*likely*, wait for treefmt's spec to be approved) be to:

1. Only read from stdin if the --stdin` option is passed.
2. The user *must* specify a filename as a positional argument.

A benefit of deprecating the implicit read from stdin behavior is users won't get confused by `nixfmt` "freezing", as reported in https://github.com/NixOS/nixfmt/issues/301.

jfly:
treefmt's stdin spec has stalled out. Our current implementation is compatible with the version of the [proposed stdin spec](https://github.com/numtide/treefmt/pull/586) that I hope will land, so I'm OK calling this done.

Issue #301, opened 2025-05-04
`nix fmt` broken


makuru-org:
`nix fmt` doesnt do anything and freezes, when used as a flake input, tested with multiple flakes.

Example flake:

```
{
  inputs = {
    naersk.url = "github:nix-community/naersk/master";
    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
    utils.url = "github:numtide/flake-utils";
  };

  outputs =
    {
      self,
      nixpkgs,
      utils,
      naersk,
    }:
    utils.lib.eachDefaultSystem (
      system:
      let
        pkgs = import nixpkgs { inherit system; };
        naersk-lib = pkgs.callPackage naersk { };
      in
      {
        formatter = pkgs.nixfmt-tree;
        defaultPackage = naersk-lib.buildPackage {
          src = ./.;
          verston = "0.1.0";
          nativeBuildInputs = with pkgs; [
            which
            cmake
            kdePackages.wrapQtAppsHook
            kdePackages.qttools
          ];
          buildInputs = with pkgs; [ kdePackages.qtbase ];
        };
        devShell =
          with pkgs;
          mkShell {
            buildInputs = [
              cargo
              rustc
              rustfmt
              pre-commit
              rustPackages.clippy
              cmake
              kdePackages.full
            ];
            RUST_SRC_PATH = rustPlatform.rustLibSrc;
          };
      }
    );
}

```

jfly:
@makuru-org, I am not able to reproduce the issue. Perhaps it's related to your locked deps? Try `nix flake update`, and if that doesn't fix things, please share a full git repo where you can repro the issue.

You could also try running `nix build .#formatter` to see if that gives you any useful output.

makuru-org:
> [@makuru-org](https://github.com/makuru-org), I am not able to reproduce the issue. Perhaps it's related to your locked deps? Try `nix flake update`, and if that doesn't fix things, please share a full git repo where you can repro the issue.
> 
> You could also try running `nix build .#formatter` to see if that gives you any useful output.

The problem, why its not reproducible is, is because I accidentally sent the one with `nixfmt-tree`,
It should have been: `formatter = pkgs.nixfmt-rfc-style;`.

jfly:
Ah, this is happening because if you pass `nixfmt` no arguments, it tries to read from stdin, which feels like a hang.

[...]

PR #358, opened 2025-11-13
Implement treefmt's stdin spec

jfly:
Full disclosure: the treefmt stdin spec is not yet finalized (see https://github.com/numtide/treefmt/pull/586). However, nothing prevents us from implementing the version of it that I hope will be approved.

This completes https://github.com/NixOS/nixfmt/issues/305.

I have very little haskell knowledge. I'm sure there are issues with this implementation. Be kind, I'm excited to learn more!

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,9 +291,9 @@ to return back to the unmerged state.
 
 ## Usage
 
-* `nixfmt < input.nix` – reads Nix code from `stdin`, formats it, and outputs to `stdout`
+* `echo "{a=1;}" | nixfmt --stdin-filepath input.nix` – reads Nix code from `stdin`, formats it, and outputs to `stdout` (the filepath (`input.nix`) is only used for error messages)
 * `nixfmt file.nix` – format the file in place
 
-## Acknowledgements
+## Acknowledgments
 
 `nixfmt` was originally developed by [Serokell](https://github.com/serokell) and later donated to become an official Nix project with the acceptance of [RFC 166](https://github.com/NixOS/rfcs/pull/166).
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -15,7 +15,6 @@ import Data.ByteString.Char8 (unpack)
 import Data.Either (lefts)
 import Data.FileEmbed
 import Data.List (intersperse, isSuffixOf)
-import Data.Maybe (fromMaybe)
 import Data.Text (Text)
 import qualified Data.Text.IO as TextIO (getContents, hPutStr, putStr)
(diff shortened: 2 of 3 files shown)


jfly (commented):
Discussed in today's team meeting. Let's preserve the old behavior (with deprecation warnings), and then sit and wait for treefmt to finalize the stdin spec.

jfly:
Discussed with @ashwinmathi. He's gonna get this PR polished up (thanks!) Notes from convo:

# TODAY
just format a file, yo:

```
nixfmt foo.nix
```

stdin mode, no "filename annotation":

```
nixfmt
```

stdin mode, with "filename annotation":

```
nixfmt --filename foo.nix
```

# Middle term

just format a file, yo:

```
nixfmt foo.nix
```

*WITH WARNING* stdin mode, no "filename annotation":

```
nixfmt
```

*WITH WARNING* stdin mode, with "filename annotation":

```
nixfmt --filename foo.nix
```

stdin mode:

```
nixfmt --stdin-filepath foo.nix
```

# Eventually

just format a file, yo:

```
nixfmt foo.nix
```

stdin mode:

```
nixfmt --stdin-filepath foo.nix
```

MattSturgeon:
Maybe this is a dumb question, but why do we need to deprecate (and eventually drop) formatting stdin without a filename?

I understand that to comply with the treefmt spec, we need to _add_ support for specifying stdin filenames, but I don't understand why we need to _drop_ support for stdin without filenames?

Surely it's fine to support things above and beyond the scope of treefmt's spec? Or is there an actual conflict I'm not aware of?

jfly:
Fair question! You are correct that we could choose to preserve the old behavior, however I personally believe the old behavior is bad UX. From #305:

> A benefit of deprecating the implicit read from stdin behavior is users won't get confused by nixfmt "freezing", as reported in https://github.com/NixOS/nixfmt/issues/301.

MattSturgeon:
Makes sense.

Some tools handle this by "detecting" stdin input; e.g. waiting a few ms for input, but continuing if nothing is received. I don't think nixfmt _needs_ that, necessary; unless someone wants to implement it.

I'd suggest we implement the treefmt spec and also `-` as a shorthand for anonymous stdin. That'd allow doing things like `echo 'nix expression' | nixfmt -` without needing a dummy filename, which would be bad UX in simple/ad-hoc cases.

jfly:
I think that's a great plan!

jfly:
Closing, as this was implemented by @ashwinmathi in #376

PR #376, opened 2026-03-17
Add deprecation warning for bare invocation of nixfmt

### LINKED ISSUE #301: `nix fmt` broken

ashwinmathi:
Added warning: `Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.`

This closes #301

### DIFF
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -169,6 +169,9 @@ checkFileTarget :: FilePath -> Target
 checkFileTarget path = Target (readFileUtf8 path) path (const $ const $ pure ())
 
 toTargets :: Nixfmt -> IO [Target]
+toTargets Nixfmt{files = [], filename = Nothing} = do
+  hPutStrLn stderr "\ESC[33mWarning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.\ESC[0m"
+  pure [stdioTarget Nothing]
 toTargets Nixfmt{files = [], filename} = pure [stdioTarget filename]
 toTargets Nixfmt{files = ["-"], filename} = pure [stdioTarget filename]
 toTargets Nixfmt{check = False, files = paths} = map fileTarget <$> collectAllNixFiles paths
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -170,7 +170,7 @@ checkFileTarget path = Target (readFileUtf8 path) path (const $ const $ pure ())
 
 toTargets :: Nixfmt -> IO [Target]
 toTargets Nixfmt{files = [], filename = Nothing} = do
-  hPutStrLn stderr "\ESC[33mWarning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.\ESC[0m"
+  hPutStrLn stderr "\ESC[33mWarning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.\ESC[39m"
   pure [stdioTarget Nothing]
 toTargets Nixfmt{files = [], filename} = pure [stdioTarget filename]
 toTargets Nixfmt{files = ["-"], filename} = pure [stdioTarget filename]


main/Main.hs
```diff
@@ -169,6 +169,9 @@ checkFileTarget :: FilePath -> Target
 checkFileTarget path = Target (readFileUtf8 path) path (const $ const $ pure ())
 
 toTargets :: Nixfmt -> IO [Target]
+toTargets Nixfmt{files = [], filename = Nothing} = do
+  hPutStrLn stderr "\ESC[33mWarning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.\ESC[0m"
```
MattSturgeon: Nit: it may be nice to reset only the color specifically, rather than resetting everything (`[39m` vs `[0m`)


```suggestion
  hPutStrLn stderr "\ESC[33mWarning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.\ESC[39m"
```

MattSturgeon: Strictly speaking, we should also avoid ANSI formatting if stderr isn't a TTY or if the `NO_COLOR`[^1] env var is set. E.g. if stderr is being redirected to a file, nixfmt is being run on a system without ANSI support, or users have set `NO_COLOR` as a preference.

I usually implement `error`, `warn`, etc logging functions (as needed), to encapsulate how different error levels should be formatted in the current environment.

IDK how this is usually done in haskell.[^2]

These are best practices, though, and probably not critical for a warning that should only appear in interactive usage.

[^1]: https://no-color.org
[^2]: https://github.com/haskell/ghcup-hs/commit/a08e62430901dcc0271fa65bd96ded68b65bd152 shows how ghcup handled `NO_COLOR`, but that's obviously more involved than nixfmt needs.

MattSturgeon (commented):
Thanks for the quick PR!

I just have a couple minor nits w.r.t. the ANSI text formatting, otherwise diff SGTM. Tested locally, and the warning appears as intended.

```
$ echo '{ foo = "bar"; }' | result/bin/nixfmt -
{ foo = "bar"; }

$ echo '{ foo = "bar"; }' | result/bin/nixfmt
Warning: Bare invocation of nixfmt is deprecated. Use 'nixfmt -' for anonymous stdin.
{ foo = "bar"; }
```

main/Main.hs
jfly: Re: NO_COLOR and tty sniffing, I'm not opposed, but this is not our first color (https://github.com/NixOS/nixfmt/blob/efdd437c3cfaf8f2d0f797d5f073e234fea815dc/main/Main.hs#L130), so I'd say let's improve that in a follow-up PR.

jfly (approved):
Thanks!

ashwinmathi:
I'll make a new PR with the error and warning stuff cleaned up, those escape codes are elsewhere in the repo I'll get em all

PR #377, opened 2026-03-20
Absorb single-line destructured params into parenthesized functions

### LINKED ISSUE #375: Reduce indentation in parentheses-wrapped functions
### Description

Functions with attribute set arguments should have the same amount of indentation compared to regular single argument functions when wrapped in a pair of parentheses.

### Small example input

```nix
{
  foo = [
    (bar: {
      baz = "foobar";
    })
    (
      { bar }:
      {
        baz = "foobar";
      }
    )
  ];
}
```

### Expected output

```nix
{
  foo = [
    (bar: {
      baz = "foobar";
    })
    ({ bar }: {
      baz = "foobar";
    })
  ];
}
```

### Actual output

```nix
{
  foo = [
    (bar: {
      baz = "foobar";
    })
    (
      { bar }:
      {
        baz = "foobar";
      }
    )
  ];
}
```

dyegoaurelio:
closes #375

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -48,6 +48,7 @@ import Nixfmt.Types (
   Expression (..),
   Item (..),
   Items (..),
+  LanguageElement (..),
   Leaf,
   ParamAttr (..),
   Parameter (..),
@@ -70,6 +71,7 @@ import Nixfmt.Types (
   tokenText,
  )
 import Nixfmt.Util (isSpaces)
+import Text.Megaparsec (Pos)
 import Prelude hiding (String)
 
 toLineComment :: TrailingComment -> Trivium
@@ -87,6 +89,10 @@ moveTrailingCommentUp a = a
 -- Prepend extra trivia to a token's existing preTrivia
 prependTrivia :: Trivia -> Ann a -> Ann a
 prependTrivia extra a@Ann{preTrivia} = a{preTrivia = extra <> preTrivia}
+
+-- | Get the source line of the first token in a language element
+firstTokenLine :: (LanguageElement a) => a -> Pos
+firstTokenLine = snd . mapFirstToken' (\a -> (a, sourceLine a))
 
 instance Pretty TrailingComment where
   pretty (TrailingComment c) =
@@ -402,17 +408,9 @@ instance Pretty Parameter where
       handleTrailingComma (x : xs) = pretty x : handleTrailingComma xs
 
       sep =
-        -- If the braces are on different lines, keep them like that
-        if sourceLine bopen /= sourceLine bclose
-          then hardline
-          else case attrs of
-            [ParamEllipsis _] -> line
-            -- Attributes must be without default
-            [ParamAttr _ Nothing _] -> line
-            [ParamAttr _ Nothing _, ParamEllipsis _] -> line
-            [ParamAttr _ Nothing _, ParamAttr _ Nothing _] -> line
-            [ParamAttr _ Nothing _, ParamAttr _ Nothing _, ParamEllipsis _] -> line
-            _ -> hardline
+        if canFlattenAttrs bopen attrs bclose
+          then line
+          else hardline
   pretty (ContextParameter param1 at param2) =
     pretty param1 <> pretty at <> pretty param2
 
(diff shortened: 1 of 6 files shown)


dyegoaurelio:
I didn't expect that the nixpkgs diff would be that big

src/Nixfmt/Pretty.hs
```diff
@@ -590,13 +579,36 @@ prettyWith _ (With with expr0 semicolon expr1) =
     <> pretty expr1
 prettyWith _ _ = error "unreachable"
 
+-- Check if a set of parameter attributes is simple enough to fit on one line:
+-- braces must be on the same source line, and up to 2 attrs without defaults,
+-- with an optional ellipsis.
```
MattSturgeon, line 611: Is it also worth considering whether the set pattern has an `@`-name?

E.g. `{ foo, bar, ... }` is fairly trivial, but `{ foo, bar, ... }@args` is less so.

And/or consider the total char count of the formatted set pattern (including `@`-name and braces), similar to how we count the total chars of a formatted line? That'd account for long arg names and long `@`-names.

src/Nixfmt/Pretty.hs
```diff
@@ -590,13 +579,36 @@ prettyWith _ (With with expr0 semicolon expr1) =
...
+    paramHasTrivia (ParamEllipsis dots) = hasTrivia dots
+
 isAbsorbableExpr :: Expression -> Bool
 isAbsorbableExpr expr = case expr of
   (Term t) | isAbsorbableTerm t -> True
   (With _ _ _ (Term t)) | isAbsorbableTerm t -> True
   -- Absorb function declarations but only those with simple parameter(s)
   (Abstraction (IDParameter _) _ (Term t)) | isAbsorbable t -> True
   (Abstraction (IDParameter _) _ body@(Abstraction{})) -> isAbsorbableExpr body
+  -- Absorb function declarations with attribute set parameters if the attribute set is simple enough, and the body is absorbable.
```
MattSturgeon, line 637: Is there a way to "soft absorb", allowing us to respect line breaks in user input?

E.g. if the user explicitly wants

```nix
{ foo, ... }:
{ }
```

to have the function body on a separate line, we could respect this.

MattSturgeon: I think this may be useful for modules, where it is common to add/remove attrs from the destructured set pattern.

Frequently switching between 2-3 attrs would cause frequent indentation change diffs.

Especially for modules defined within a wider expression, e.g.:
```nix
{
  perSystem =
    { pkgs, system, ... }:
    { };
}
```

If we start adding/removing a third argument from that example, the entire function body gets indented/dedented.

MattSturgeon: This would also avoid a treewide nixpkgs diff, since we already have the line breaks in nixpkgs.

Instead, we'd have the option of looking for functions that benefit from absorbed formatting and manually removing their line breaks.

MattSturgeon (commented):
Thanks for taking the time to implement this!

I still think it's generally the right direction. However, I'm also a little cautious of the implications for diff noise resulting from indentation changing back and forth, which I hadn't considered in the last meeting.

test/diff/lambda/out-pure.nix
```diff
@@ -88,21 +85,13 @@ in
...
   )
 
-  (
-    a:
-    { b, ... }:
-    c: {
-      # Stuff
-    }
-  )
+  (a: { b, ... }: c: {
```
piegamesde, line 88: Is this really an improvement? IMO, mixed plain and attrs function args should not be on the same line

test/diff/lambda/out-pure.nix
```diff
@@ -200,23 +189,110 @@ in
...
+    {
+      a = 1;
+      b = 2;
+    }
+  )
+
+  # comment on attr parameter
+  (
+    # comment
+    { a }: { x = 1; })
```
piegamesde: Closing parenthesis is wrong here

src/Nixfmt/Pretty.hs
dyegoaurelio: this would be tricky to implement. We don't have the concept of soft absorption in the formatter. 

Another alternative would be to don't mark the expression as absorbable if the source line is the same as it's parent.

However, given this input
```nix
    perSystem =
      { pkgs, system, ... }:
      {
        out = pkgs.lib.optionalString (system == "x86_64-linux") "foo";
      };
    
    perSystem = { pkgs, system, ... }:
      {
        out = pkgs.lib.optionalString (system == "x86_64-linux") "foo";
      };
```

the goal would be to **not** absorb the first one, but absorb the second one, right?

The problem is that for both cases we'd be resolving  whether the expression
```nix
{ pkgs, system, ... }:
      {
        out = pkgs.lib.optionalString (system == "x86_64-linux") "foo";
      };
``` 
is absorbable. (it has no context about where it's positioned)

dyegoaurelio: So in short, it's possible but would require much more changes to the code

jfly (approved):
Team-reviewed in this week's team meeting. LGTU! Looking forward to it =)

PR #381, opened 2026-04-10
feat: Update project, actions, Haskell format, and release more builds to GitHub release to support CI workflows

RobertDeRose:
This PR is structured into a series of logically scoped commits to enable the GitHub release workflow to produce additional build artifacts. The goal is to allow tools such as `mise-en-place` and `hk` to install `nixfmt` without requiring a full Nix environment, particularly in constrained environments like GitHub Actions runners.

A significant portion of the work focused on producing static builds for both `x86_64` and `arm64`. This led to updates to `nixpkgs` and the removal of the `treefmt` patch. Despite these efforts, GHC does not appear to reliably support static builds on `arm64`. Given the limits of my Haskell experience, I stopped further investigation at that point.

As a pragmatic alternative, this PR expands support across modern Linux and Darwin systems rather than pursuing full static portability.

The release workflow has been refactored to improve clarity and efficiency:
* Version detection now runs as a separate job. If no new version is found, the workflow exits early. This eliminates the need for conditional logic in individual steps.
* Builds are executed via a matrix across supported operating systems and architectures, avoiding redundant checks per environment.
* The release process is isolated into its own job. A release is only created if all builds succeed, and all generated artifacts are attached at creation time.
* Additional, the release workflow can be triggered manually using a `workflow_dispatch` which will use the latest version number and build a pre-release with the tag being the `${version}-test`, i.e. `v1.2.0-test`

Release artifacts are named by operating system and architecture. This enables tools like `mise` to resolve and install the correct binary automatically (e.g., `mise use github:NixOS/nixfmt`).

Validation was performed on **Ubuntu Jammy** and **Noble**, and **macOS 15** and **26**, across both **arm64** and **x86_64**. All unit tests pass, and local builds succeed using `nix develop` followed by `cabal build`.

Given the scope of changes, commits have been kept intentionally granular so that individual changes can be reviewed, modified, or dropped as needed.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -160,6 +160,8 @@ jobs:
           if [[ "$IS_PRERELEASE" == "true" ]]; then
             TAG="v${VERSION}-test"
             PRERELEASE_FLAG="--prerelease"
+            # Delete existing pre-release if it exists
+            gh release delete "$TAG" --yes --cleanup-tag 2>/dev/null || true
           else
             TAG="v${VERSION}"
             PRERELEASE_FLAG=""
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,6 +6,7 @@ on:
       - master
     paths:
       - 'nixfmt.cabal'
+  workflow_dispatch:
 
 
 defaults:
@@ -23,6 +24,7 @@ jobs:
       prev_version: ${{ steps.get_version.outputs.prev_version }}
       version: ${{ steps.get_version.outputs.version }}
       version_changed: ${{ steps.get_version.outputs.version_changed }}
+      is_prerelease: ${{ steps.get_version.outputs.is_prerelease }}
 
     steps:
       - name: Checkout main
@@ -41,6 +43,14 @@ jobs:
           echo "Version found in cabal: $VERSION"
           echo "version=$VERSION" >> $GITHUB_OUTPUT
 
+          if [[ "${{ github.event_name }}" == "workflow_dispatch" ]]; then
+            echo "Manual dispatch — skipping version comparison"
+            echo "prev_version=0" >> $GITHUB_OUTPUT
+            echo "version_changed=true" >> $GITHUB_OUTPUT
+            echo "is_prerelease=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
           # Get previous version from last commit
           PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}')
           echo "Previous version: $PREV_VERSION"
@@ -49,9 +59,11 @@ jobs:
 
           if [ "$PREV_VERSION" != "$VERSION" ]; then
             echo "version_changed=true" >> $GITHUB_OUTPUT
+            echo "is_prerelease=false" >> $GITHUB_OUTPUT
             echo "::notice ::✅ Version changed from $PREV_VERSION to $VERSION"
           else
             echo "version_changed=false" >> $GITHUB_OUTPUT
+            echo "is_prerelease=false" >> $GITHUB_OUTPUT
             echo "::notice ::Version did not change"
           fi
 
@@ -109,6 +121,7 @@ jobs:
       contents: write
     env:
       VERSION: ${{ needs.check-if-version-changed.outputs.version }}
+      IS_PRERELEASE: ${{ needs.check-if-version-changed.outputs.is_prerelease }}
     steps:
       - name: Checkout main
         uses: actions/checkout@v6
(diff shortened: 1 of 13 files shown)


dyegoaurelio:
Thanks for the contribution.

Can you explain why these changes are necessary and why the mentioned tools can't execute the static binaries we already provide? e.g https://github.com/NixOS/nixfmt/releases/download/v1.2.0/nixfmt

Edit:
looks like what you're trying to accomplish is having arm64 and darwin binaries, right?

RobertDeRose:
@dyegoaurelio — I may be off base, but my goal is to run `nixfmt` without requiring Nix to be installed, particularly on macOS (Intel and arm64).

I’m using `nix-darwin` and `numtide:system-manager`, with a bootstrap script that leverages `mise-en-place` to install linters and pre-commit hooks for `hk`, including `nixfmt`. This works when Nix is already installed, but fails otherwise.

The current static build is limited to x86_64 Linux, which doesn’t help in my case since my Linux systems are arm64. Running it on macOS would require emulation (e.g., binfmt or Rosetta), which feels like the wrong approach for a tool like this. It should be straightforward to use across platforms.

My thinking behind this PR is that once users have Nix set up, they don’t rely on these builds anyway. So providing only a single x86_64 Linux static binary has limited value.

Instead, it seems more useful to support a broader set of targets. That’s the motivation for this PR.

This started as an attempt to get static arm64 builds working, but the Haskell toolchain (at least from my limited experience) doesn’t seem to support that well. The additional changes were required in that effort; while the goal didn’t pan out, they still appeared to me to add value worth keeping.

jfly (commented):
Discussed in today's team meeting. @RobertDeRose, thanks again for the contributions.

We're OK with providing static builds of nixfmt for more platforms than we already do, but we want to continue to use nix to do those builds.

Could you separate out your other changes into separate PRs? These 2 changes seem (almost) ready to go:

- "Update nixpkgs pin to nixos-25.11 and remove obsolete treefmt patch". But please keep us on the unstable branch instead.
- "Pass '-' to nixfmt in tests to avoid bare invocation warning"

RobertDeRose:
@jfly So I made a PR based on just the two commits you suggested, but that breaks the builds because now I need a lot of the other changes I made to support the builds.

How would you feel if I just revert most of the release workflow to just not include linux/arm64 static builds? Darwin already doesn't support static builds, so nothing can change there anyway.

I honestly think that this PR solves the problem properly as is, since it's really just meant to help with things like CI workflows to be able to grab a build that works on Ubuntu or macOS without needing to install an entire nix environment and store a cache for everything.

Perhaps I'm just doing my CI wrong though, I am a complete noob when it comes to Nix, I just started working on it in December on the as a hobby.

I honestly just started down this rabbit hole because I wanted a formatter before I had a working nix setup :)

jfly:
@RobertDeRose, please just put together minimal, functional PRs. If it's easier for you to evolve this PR into one by removing commits, that's fine!

> I honestly think that this PR solves the problem properly as is

What does "the problem" refer to here?

.github/workflows/release.yml
```diff
@@ -29,74 +41,134 @@ jobs:
...
+          cabal configure \
+            --enable-executable-stripping \
+            --enable-optimization=2
+          cabal build exe:nixfmt
+
+          name="nixfmt_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m)"
+
+          bin=$(cabal list-bin nixfmt)
+          strip "$bin"
+          cp "$bin" "${name}"
```
dyegoaurelio, line 108: aside from the unrelated changes that deserve their own pr, this is the biggest issue with the approach taken.

All the target platforms are supported by nix, I don't see why we can't use nix to build these binaries on all these hosts.

Also, we're already providing a static binary for linux_amd64, so we should keep it to avoid breaking existing scripts.

the result should be something like this

- `nixfmt` (alias to `nixfmt_linux_amd64`)
-  `nixfmt_linux_amd64`
-  `nixfmt_linux_arm64`
-  `nixfmt_darwin_amd64`
- `nixfmt_darwin_arm64`

dyegoaurelio: Also, the provided binaries **shouldn't be dynamically** linked. They should be static binaries (so they don't depend on the user having specific libraries available)

```
[dyego@nixos-desktop:/tmp]$ curl -LO https://github.com/NixOS/nixfmt/releases/download/v1.2.0/nixfmt
  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed
  0      0   0      0   0      0      0      0                              0
100  5.98M 100  5.98M   0      0  6.48M      0                              0

[dyego@nixos-desktop:/tmp]$ ldd ./nixfmt
ldd: warning: you do not have execution permission for `./nixfmt'
        not a dynamic executable

[dyego@nixos-desktop:/tmp]$ curl -LO https://github.com/RobertDeRose/nixfmt/releases/download/v1.2.0-test/nixfmt_linux_x86_64
  % Total    % Received % Xferd  Average Speed  Time    Time    Time   Current
                                 Dload  Upload  Total   Spent   Left   Speed
  0      0   0      0   0      0      0      0                              0
100  7.21M 100  7.21M   0      0  5.59M      0   00:01   00:01              0

[dyego@nixos-desktop:/tmp]$ ldd ./nixfmt_linux_x86_64
ldd: warning: you do not have execution permission for `./nixfmt_linux_x86_64'
        linux-vdso.so.1 (0x00007f7f428da000)
        libm.so.6 => /nix/store/pdwlyjkmc6icc0wb5gaw08m9ynqrg683-glibc-2.40-218/lib/libm.so.6 (0x00007f7f427ea000)
        libgmp.so.10 => not found
        libc.so.6 => /nix/store/pdwlyjkmc6icc0wb5gaw08m9ynqrg683-glibc-2.40-218/lib/libc.so.6 (0x00007f7f42400000)
        /lib64/ld-linux-x86-64.so.2 => /nix/store/pdwlyjkmc6icc0wb5gaw08m9ynqrg683-glibc-2.40-218/lib64/ld-linux-x86-64.so.2 (0x00007f7f428dc000)

[dyego@nixos-desktop:/tmp]$
```

jfly:
Closing this. We're OK with providing more statically build binaries, but we'd want the builds to occur with idiomatic usage of nix itself.

PR #382, opened 2026-04-10
Add Dependabot for nix flake updates

JamieMagee:
See https://github.blog/changelog/2026-04-07-dependabot-version-updates-now-support-the-nix-ecosystem/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "nix"
+    directory: "/"
+    schedule:
+      interval: "weekly"


dyegoaurelio (approved):
Nice! I wasn't aware about this update

dyegoaurelio:
actually, this project saves it's dependencies on `npins/sources.json`
I don't think dependabot will caught that.

We would only get prs for `@numtide/flake-utils` and `@nix-systems/default` updates

JamieMagee:
﻿Good point. The flake.lock only has flake-utils and systems, so this doesn't cover much. The real dependencies are in npins. Happy to close this if it's not worth the noise for two inputs.

jfly (approved):
Thanks! We'll figure out the npins stuff some other time.

PR #386, opened 2026-05-04
test: add three failing --verify regressions found by fuzzing

Mic92:
All three inputs make `nixfmt --verify` report a bug on current master; the `out.nix` files encode the fixed point that a corrected formatter should reach.

- idempotent_lang_annotation
    A trailing `# x` on `[` is rendered on its own line, where it
    re-lexes as a leading comment merged into the same trivia run as the
    following `/* sh */` language annotation. Pass 1 fuses the
    annotation with the string, pass 2 sees `[LineComment, LangAnn]` and
    splits them.

- idempotent_leading_blank
    Leading blank lines on the file's first token make the head term
    fail `isSimple`, so `prettyApp` takes the non-simple branch on
    pass 1. The output has no leading blanks, so pass 2 takes the simple
    branch and absorbs the argument.

- idempotent_div_lang_annotation
    `a / /* sh */ "" p` is emitted as `a //* sh */ "" p`: the hoisted
    language-annotation comment lands directly after the `/` with no
    spacing, and `//` re-lexes as the update operator, so the output
    fails to parse.


test/correct/integer-selection.nix
```diff
@@ -0,0 +1,8 @@
+# Selecting an attribute on a numeric literal must keep a separating space:
+# `1.a` re-lexes as the float `1.` applied to `a`, which is a different
+# program. The space below is therefore load-bearing.
+{
+  a = 1 .a;
+  b = 42 .foo.bar or 0;
```
Mic92, line 6: this gets a bit crazy... but I have those fixed in nixfmt, so it doesn't crash the in the fuzzer.

test/diff/idempotent_multiline_string_trailing_comment/out-pure.nix
```diff
@@ -0,0 +1,6 @@
+[
+  "
+"
+  # c
```
Mic92: Those are tricky to fix with the current AST...

Mic92:
I will add in the meantime vendor these patches to address idempotency in my tests: https://github.com/Mic92/nixfmt/pull/new/idempotency-fixes

dyegoaurelio:
thanks for finding these issues.

I just created #387  to address them

dyegoaurelio:
Sorry, I accidentally pushed to you branch.

It should be back where it was

PR #387, opened 2026-05-06
Fix idempotency regressions found on #386 

dyegoaurelio:


MattSturgeon (approved):
Thanks @dyegoaurelio and @Mic92

PR #389, opened 2026-05-12
Update nixpkgs and various associated fixes

jfly:
Note: There are multiple commits in here, I suggest reading them one at a time.

PR #390, opened 2026-05-17
Enforce that there are no warnings in our haskell code

jfly:
(There are multiple commits in here, I suggest reading them separately)

We had a bunch of warnings enabled, but the build would still succeed if
any of them fired. Now we have a project wide `werror` flag which
determines whether or not these warnings are fatal (IIUC, this is best
practice in the haskell ecosystem, as others may want to build your
project with a different version of ghc, which may emit different
warnings.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,7 +11,8 @@ You can also reach us on Matrix at `#nix-formatting:nixos.org`.
 
 Enter a development shell with `nix-shell`, `nix develop` or automatically with [direnv](https://direnv.net/), after which you can:
 - Build (& run): `cabal build` (`cabal run`)
-- Build with fatal warnings (as in CI): `cabal build --flag werror`. May require `cabal clean` beforehand.
+- Build and ignore warnings: `cabal build --flag -werror`. This is useful when
+  hacking locally, but code must be free of warnings before merging.
 - Debug: `cabal repl`
 - Format the codebase: `treefmt`
 - [Set up](https://haskell-language-server.readthedocs.io/en/latest/configuration.html#configuring-your-editor) your LSP-editor to use `haskell-language-server`
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,6 +11,7 @@ You can also reach us on Matrix at `#nix-formatting:nixos.org`.
 
 Enter a development shell with `nix-shell`, `nix develop` or automatically with [direnv](https://direnv.net/), after which you can:
 - Build (& run): `cabal build` (`cabal run`)
+- Build with fatal warnings (as in CI): `cabal build --flag werror`. May require `cabal clean` beforehand.
 - Debug: `cabal repl`
 - Format the codebase: `treefmt`
 - [Set up](https://haskell-language-server.readthedocs.io/en/latest/configuration.html#configuring-your-editor) your LSP-editor to use `haskell-language-server`
--- a/nixfmt.cabal
+++ b/nixfmt.cabal
@@ -19,6 +19,11 @@ source-repository head
   type:     git
   location: git://github.com/NixOS/nixfmt.git
 
+flag werror
+  description: Treat warnings as errors
+  default: False
+  manual: True
+
 common warnings
   ghc-options:
     -Wall
@@ -27,6 +32,10 @@ common warnings
     -Wincomplete-uni-patterns
     -Wredundant-constraints
     -Wno-orphans
+
+  if flag(werror)
+    ghc-options:
+      -Werror
 
 executable nixfmt
   import: warnings
--- a/nixfmt.cabal
+++ b/nixfmt.cabal
@@ -1,4 +1,4 @@
-cabal-version:       2.0
+cabal-version:       3.8
 name:                nixfmt
 version:             1.2.0
 synopsis:            Official formatter for Nix code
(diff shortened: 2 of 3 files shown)


CONTRIBUTING.md
```diff
@@ -11,6 +11,7 @@ You can also reach us on Matrix at `#nix-formatting:nixos.org`.
 
 Enter a development shell with `nix-shell`, `nix develop` or automatically with [direnv](https://direnv.net/), after which you can:
 - Build (& run): `cabal build` (`cabal run`)
+- Build with fatal warnings (as in CI): `cabal build --flag werror`. May require `cabal clean` beforehand.
```
MattSturgeon: Is there a way to make fatal warnings the default, so that we don't need to explicitly mention this or build in a special way?

nixfmt.cabal
```diff
@@ -19,7 +19,27 @@ source-repository head
   type:     git
   location: git://github.com/NixOS/nixfmt.git
 
+flag werror
+  description: Treat warnings as errors
+  default: False
```
MattSturgeon: E.g. would setting `default: True` here make sense?

nixfmt.cabal
```diff
@@ -19,7 +19,27 @@ source-repository head
...
+    -Wall
+    -Wcompat
+    -Wincomplete-record-updates
+    -Wincomplete-uni-patterns
+    -Wredundant-constraints
+    -Wno-orphans
+
+  if flag(werror)
+    ghc-options:
+      -Werror
```
MattSturgeon, line 38: Q: does this extend the above options, or replace them?

nixfmt.cabal
jfly: This does work! I originally avoided it because `cabal check` doesn't like it:

```
Warning: 'ghc-options: -Werror' makes the package easy to break with future
GHC versions because new GHC versions often add new warnings. Alternatively,
if you want to use this, make it conditional based on a Cabal configuration
flag (with 'manual: True' and 'default: False') and enable that flag during
development.
```

IMO, the reasoning here is about haskell libraries being published to hackage, which doesn't apply to us. I do add a check that `cabal check` is happy in https://github.com/NixOS/nixfmt/pull/389, so I'll have to tweak that if we make this change.

WDYT?

MattSturgeon: It'd also apply to us when bumping our GHC version (via bumping nixpkgs). But I still feel like if we want warnings to act like errors, we should probably set `-Werror` by default.

Not publishing to hackage does at least mean we only ever need to target/support _one_ GHC version at a time.

nixfmt.cabal
jfly: Good question. It extends them. Proof:

Add a warning:
```diff
$ git diff
diff --git a/main/Main.hs b/main/Main.hs
index 6ec11ae..39eece2 100644
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -248,6 +248,9 @@ writeErrorBundle doWrite result = do
 runJobs :: (String -> IO ()) -> [IO Result] -> IO [Result]
 runJobs writeError = mapM (>>= writeErrorBundle writeError)
 
+f :: (Eq a, Ord a) => a -> String
+f _ = "hiya"
+
 main :: IO ()
 main = withUtf8StdHandles $ do
   _ <-
```

It fails to build:

```console
$ cabal build --flag werror
...
main/Main.hs:251:6: error: [-Wredundant-constraints, -Werror=redundant-constraints]
    Redundant constraints: (Eq a, Ord a)
    In the type signature for:
         f :: forall a. (Eq a, Ord a) => a -> String
    |
251 | f :: (Eq a, Ord a) => a -> String
    |      ^^^^^^^^^^^^^
Error: cabal: Failed to build exe:nixfmt from nixfmt-1.2.0.
```

Now remove `-Wredundant-constraints` from the list of `ghc-options`:

```diff
diff --git a/nixfmt.cabal b/nixfmt.cabal
index fa39023..bd9358c 100644
--- a/nixfmt.cabal
+++ b/nixfmt.cabal
@@ -30,7 +30,6 @@ common warnings
     -Wcompat
     -Wincomplete-record-updates
     -Wincomplete-uni-patterns
-    -Wredundant-constraints
     -Wno-orphans
 
   if flag(werror)
```

The build now succeeds:

```console
$ cabal build --flag werror
$ echo $?
0
```

CONTRIBUTING.md
jfly: There is. Marking this resolved, let's continue this conversation in https://github.com/NixOS/nixfmt/pull/390#discussion_r3265410542.

dyegoaurelio: We could add a flake check instead of recommending ad hoc commands

nixfmt.cabal
dyegoaurelio: I personally find it a bit annoying when the compiler refuses to compile due to a warning.

But I agree that having it as the default is probably easier for new contributors.

So I'm fine either way

CONTRIBUTING.md
jfly: It is part of our flake check (the checks in `default.nix` get exposed as flake checks).

I wrote it down because it took me a little while to figure out, and it seemed to fit with the fact that we document `cabal run` here.

nixfmt.cabal
jfly: > But I agree that having it as the default is probably easier for new contributors.

I agree. I've changed the default to enable the `werror` flag, and documented how to turn it off for folks who are quickly hacking on stuff.

CONTRIBUTING.md
jfly: Marking resolved. The default is now to be strict (enable the `werror` flag), and this documentation tells you how to turn it off if you just want to do something quick-n-dirty.

dyegoaurelio (approved):

PR #391, opened 2026-05-19
ci: Update and pin 3rd party GitHub actions

LeSuisse:
This makes sure the first level of actions pulled from the workflows cannot be replaced without us knowing about it.

See for example this past incident:
https://unit42.paloaltonetworks.com/github-actions-supply-chain-attack/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,3 +4,7 @@ updates:
     directory: "/"
     schedule:
       interval: "weekly"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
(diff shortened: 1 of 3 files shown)


LeSuisse:
Side note: I did not added the update of the actions to the Dependabot config but this is probably something that should be considered.

MattSturgeon:
> Side note: I did not added the update of the actions to the Dependabot config but this is probably something that should be considered.

Definitely. As this has a merge conflict now, would you be willing to add that when rebasing? Alternatively I can handle it separately.

MattSturgeon (approved):
Thanks

PR #392, opened 2026-05-19
ci: Do not persist checkout credentials

LeSuisse:
These credentials are never re-used so there is no good reason for keeping them.

Note: More recent versions of actions/checkout improved a bit the behavior, see https://github.com/orgs/community/discussions/179107

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - uses: actions/checkout@v4
         if: github.event_name != 'pull_request_target'
+        with:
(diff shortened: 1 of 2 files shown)


MattSturgeon (approved):
Thanks!

> Note: More recent versions of actions/checkout improved a bit the behavior,

We're still using checkout v4 👀  ?! Should definitely look into dependabot (and maybe SHA-pinning actions).

LeSuisse:
Yep :)

Upgrade and pinning available has been done here: https://github.com/NixOS/nixfmt/pull/391

PR #393, opened 2026-05-20
ci: Refine permissions

LeSuisse:
No need to request credentials/permissions that are not absolutely needed.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -10,7 +10,6 @@ on:
       - master
 
 permissions:
-  pull-requests: write
   contents: read
 
 jobs:
@@ -50,6 +49,9 @@ jobs:
     concurrency:
       group: ${{ github.workflow_ref }}-${{ github.event.pull_request.number }}
       cancel-in-progress: true
+    permissions:
+      contents: read
+      pull-requests: write
     steps:
       - name: Find Comment
         uses: peter-evans/find-comment@b30e6a3c0ed37e7c023ccd3f1db5c6c0b0c23aad # v4.0.0
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -4,6 +4,8 @@ on:
   # because we need repository secrets for that, which pull_request doesn't allow from forks.
   # However, it's very important that we don't run code from forks without sandboxing it,
   # because that way anybody could potentially extract repository secrets!
+  # We must ensure that no secrets are in any environment variables when running untrusted code, and that no secrets are persisted to disk and accessible to `readFile`
+  # when evaluating Nix code
   pull_request_target:
   push:
     branches:
@@ -27,14 +29,14 @@ jobs:
           # To prevent running untrusted code from forks,
           # pull_request_target will cause the base branch to be checked out, not the PR branch.
           # In our case we check out the PR branch regardless,
-          # because we're sandboxing all untrusted code with a `nix-build`.
-          # (and the sandbox is enabled by default at least on Linux)
+          # as we do not expose secrets when calling untrusted code
           ref: refs/pull/${{ github.event.pull_request.number }}/merge
           persist-credentials: false
 
       - uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3 # v31.10.6
 
       - uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
+        if: github.event_name != 'pull_request_target'
         with:
           name: nixos-nixfmt
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -82,11 +85,6 @@ jobs:
 
       - uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3 # v31.10.6
 
-      - uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
-        with:
-          name: nixos-nixfmt
-          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-
```
MattSturgeon, line 89: It'd be nice to keep this on `push` events, or even have it without the `authToken` on `pull_request_target` events.

Something like
```yaml
authToken: ${{ github.event_name == 'push' && secrets.CACHIX_AUTH_TOKEN || '' }}
```

.github/workflows/main.yml
MattSturgeon, line 6: Is it worth extending this comment with a note about the eval itself still being partially untrusted?

E.g. something to the effect of

> ensure `getEnv` is not allowed and no secrets are persisted to disk and accessible to `readFile`

MattSturgeon: Another potentially misleading note here

MattSturgeon: Maybe also here?

MattSturgeon (approved):
Thanks. LMK what you think of the optional feedback; whether you agree or disagree and whether you want to address any of it in this PR or not.

LeSuisse:
Made some adjustments to the comment, tell me if it looks OK.

@mdaniels5757 I added myself as co-author to take responsibility if someone needs to be yelled at in the future if the comments are wrong :sweat_smile:

jfly (approved):
LGTM, but will wait for @MattSturgeon to take another look before merging.

.github/workflows/main.yml
jfly: Non-blocking thoughts:

- How does [magic-nix-cache](https://github.com/DeterminateSystems/magic-nix-cache) handle this?
- Does Cachix have a recommendation about how to do this safely? It doesn't seem crazy to want evolving PRs to avoid rebuilds.

LeSuisse: * It uses similar semantics than the GitHub Actions cache see https://docs.github.com/en/actions/reference/workflows-and-actions/dependency-caching#restrictions-for-accessing-a-cache
* Not sure there is a great way to handle it, maybe you can have a dedicated untrusted cache for PRs with the trusted cache configured as the upstream cache. But even with something like that it's bit weird as you need to maintain a long lived write token that most likely is going to leak at some point.

.github/workflows/main.yml
```diff
@@ -4,13 +4,14 @@ on:
   # because we need repository secrets for that, which pull_request doesn't allow from forks.
   # However, it's very important that we don't run code from forks without sandboxing it,
   # because that way anybody could potentially extract repository secrets!
+  # Ensure `getEnv` is not allowed and no secrets are persisted to disk and accessible to `readFile`
```
jfly: @LeSuisse, have we done anything to prevent access to `getEnv`? AFAICT, the answer is "no".

MattSturgeon: I think my suggested phrasing was based on the assumption that environment variables will always contain some secrets. If that's not the case, then it's ok for eval to have access to `getEnv`, so long as we ensure no secrets are accessible on the environment.

The alternative is to switch from impure classic commands (`nix-build`) to the new purer commands (`nix build`). That seems like scope creep that may be best handled in a separate PR.

The wording in the other comment below handles it better:

> Make sure you do not expose secrets in the environment or persist them to disk

.github/workflows/main.yml
MattSturgeon: My preference is still to keep cachix-action here, but only providing `authToken` in trusted contexts (like `push` events).

IIRC, cachix-action will continue to function read-only when configured without a token.

If debate will delay this PR unnecessarily, then I'm happy to deal with that in a dedicated "restore cachix" PR.

.github/workflows/main.yml
mdaniels5757:
```suggestion
  # We must ensure that no secrets are in any environment variables when running untrusted code, and that no secrets are persisted to disk and accessible to `readFile`
```

.github/workflows/main.yml
jfly: > If debate will delay this PR unnecessarily, then I'm happy to deal with that in a dedicated "restore cachix" PR.

It will not! Let's discuss in https://github.com/NixOS/nixfmt/issues/394

jfly: I'm going to merge this as is. @MattSturgeon, feel free to tweak the logic so we (safely) push more things to the cache.

.github/workflows/main.yml
mdaniels5757: At present, they do not contain any secrets.

PR #398, opened 2026-05-21
workflows/main: re-enable read-only cachix for untrusted builds

MattSturgeon:
This partially reverts 903898a5e850e4b6d4453b93b51caec19370e39f from #393.

Cachix is configured on all runs, but the `authToken` is only configured on `push` events to avoid exposing it to untrusted PR code.

cc @mdaniels5757

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -36,10 +36,10 @@ jobs:
       - uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3 # v31.10.6
 
       - uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
-        if: github.event_name != 'pull_request_target'
         with:
           name: nixos-nixfmt
-          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          authToken: ${{ github.event_name == 'push' && secrets.CACHIX_AUTH_TOKEN || '' }}
+          skipPush: ${{ github.event_name != 'push' }}
 
       - name: checks
         run: nix-build -A ci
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -36,10 +36,10 @@ jobs:
       - uses: cachix/install-nix-action@8aa03977d8d733052d78f4e008a241fd1dbf36b3 # v31.10.6
 
       - uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
-        if: github.event_name != 'pull_request_target'
         with:
           name: nixos-nixfmt
-          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          authToken: ${{ github.event_name == 'push' && secrets.CACHIX_AUTH_TOKEN || '' }}
+          skipPush: ${{ github.event_name != 'push' }}
```
jfly, line 42: The fact that this boolean expression is not exactly the opposite of the expression for `authToken` makes my brain itch.

Do we need this here because an empty `authToken` means something different than no auth token at all? My reading of the relevant code is that setting `authToken` to empty string will cause us to skip pushes (https://github.com/cachix/cachix-action/blob/5f2d7c5294214f71b873db4b969586b980625e71/src/main.ts#L113).

Or is this just here to be explicit? If so, that seems fine to me, but I'd consider making the expression the boolean opposite of the`authToken` one just to reduce this "brain itch" it's giving me.

MattSturgeon: > Or is this just here to be explicit?

Yes. An empty/missing `authToken` does mean we can't write to cachix. But I wasn't sure if not setting `skipPush` explicitly would _fail_ to push instead of gracefully _skipping_ the push.

jfly: Reading the code, it looks like it would be ok to omit `skipPush`. However, leaving it here to be explicit sounds fine to me, I would update the boolean logic so they're perfect inverses of each other.

jfly: Discussed on matrix. tl;dr: the intent is not for these boolean expressions to be opposites, it's for `skipPush` to say "on pull requests (or non-pushes) where we explicitly avoid defining authToken, don't bother trying to push to cachix". This PR achieves that.

It's also all a little redundant given the [underlying code](https://github.com/cachix/cachix-action/blob/5f2d7c5294214f71b873db4b969586b980625e71/src/main.ts#L113), but I like the thought of being explicit.

jfly (approved):

PR #399, opened 2026-05-21
ci: Fix remaining issues spotted by Zizmor

LeSuisse:
A scan using [Zizmor](https://docs.zizmor.sh/) with the default persona now comes clean.

Permissions of the release workflow have been limited to only what's needed.

Once the treefmt-nix pin is updated to something more recent it will be possible to activate Zizmor (`programs.zizmor.enable = true;`) in the Treefmt config (and so have it verified during CI checks).

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -6,7 +6,7 @@ on:
   # because that way anybody could potentially extract repository secrets!
   # We must ensure that no secrets are in any environment variables when running untrusted code, and that no secrets are persisted to disk and accessible to `readFile`
   # when evaluating Nix code
-  pull_request_target:
+  pull_request_target: # zizmor: ignore[dangerous-triggers]
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


jfly (approved):
Thanks!

> Once the treefmt-nix pin is updated to something more recent it will be possible to activate Zizmor (programs.zizmor.enable = true;) in the Treefmt config (and so have it verified during CI checks).

We're getting close to this (https://github.com/NixOS/nixfmt/pull/389). Do you mind sending in a PR that adds this? It'll fail right now, but serve as a reminder to enable this once the updates have landed.

MattSturgeon:
We could also use nixpkgs' `treefmt.withConfig` and configure tools like zizmor and actionlint ourselves, but it is nice that treefmt-nix comes with sane defaults.

EDIT: seems our pinned nixpkgs pre-dates `treefmt.withConfig` anyway

PR #404, opened 2026-05-22
{flake,cabal}: update descriptions, drop `apps`

MattSturgeon:
Small followup to #401.

New descriptions are based on the first line of README.md

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -1,4 +1,6 @@
 {
+  description = "The official formatter for Nix language code";
+
   inputs.systems = {
     type = "github";
     owner = "nix-systems";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -31,6 +33,7 @@
       apps = mapResults (result: {
         default.type = "app";
         default.program = "${result}/bin/nixfmt";
+        default.meta.description = "Run nixfmt";
```
MattSturgeon: This may not be enough justification to keep the `apps` output.

I'm happy to drop the `apps` output if you'd rather just have `nix run` use `packages.default`.

jfly: Your call.

jfly (approved):

flake.nix
MattSturgeon: I'll drop it and merge.

PR #406, opened 2026-05-25
serokell-nix: drop dependency

### LINKED ISSUE #405: Upgrade to the latest version of serokell-nix
We use serokell-nix's overlay here: https://github.com/NixOS/nixfmt/blob/5b1588d898d9d9d50652461494a6eda66a7f415d/main.nix#L21

It provides hlint, which we use here: https://github.com/NixOS/nixfmt/blob/5b1588d898d9d9d50652461494a6eda66a7f415d/main.nix#L101

However, the way we import the overlay (`import (serokell-nix + "/overlay")`) is not compatible with the latest version of serokell-nix. In https://github.com/serokell/serokell.nix/commit/1e48c18d74c5505d057ff298f6e0940cd9119431, they added a `{ inputs }` parameter to `‎overlay/default.nix‎`.

I see 4 options to fix this:

1. This `inputs` parameter is used for `weeder`, which we don't use, so we could fix (workaround) this issue with this simple hack:

   ```diff
   diff --git a/main.nix b/main.nix
   index 6cc4282..e994ad7 100644
   --- a/main.nix
   +++ b/main.nix
   @@ -18,7 +18,7 @@ let
        inherit system;
        overlays = [
          overlay
   -      (import (serokell-nix + "/overlay"))
   +      (import (serokell-nix + "/overlay") { inputs = { }; })
        ];
        config = { };
      };
   ```
2. serokell-nix provides a [default.nix that uses flake-compat](https://github.com/serokell/serokell.nix/blob/master/default.nix). This feels "more correct", at the cost of us now having 2 pinned versions of nixpkgs floating around.
3. We could switch our project to be flake-first, and consume serokell-nix as a regular flake input. We can then use input follows to control how many nixpkgs we have floating around.
4. We could drop serokell-nix. The `hlint` helper is [totally trivial](https://github.com/serokell/serokell.nix/blob/master/overlay/default.nix#L79-L82).

jfly:
This fixes https://github.com/NixOS/nixfmt/issues/405

Many thanks for Serokell for starting and donating this project!


main.nix
```diff
@@ -94,7 +92,16 @@ let
       ];
       installPhase = "touch $out";
     };
-    hlint = pkgs.build.haskell.hlint haskellSource;
+    hlint =
+      pkgs.runCommand "hlint"
```
MattSturgeon, line 96: Is this trivial\* enough to justify `runCommandLocal`?

_\*and/or fast and/or unique_

MattSturgeon (approved):

main.nix
jfly: I don't know :(. I personally don't like making these decisions, I'd rather nix have its own heuristics for this sort of stuff.

jfly: I'm going to merge this as is, but feel free to change it if you want!

MattSturgeon: `runCommandLocal` (or passing `runLocal = true` to `runCommandWith`) configures [`preferLocalBuild`](https://nix.dev/manual/nix/2.34/language/advanced-attributes.html#adv-attr-preferLocalBuild) and [`allowSubstitutes`](https://nix.dev/manual/nix/2.34/language/advanced-attributes.html#adv-attr-allowSubstitutes) to instruct `nix` to skip network requests for remote builders and binary substituters ([`runCommandWith` impl](https://github.com/NixOS/nixpkgs/blob/282deb2d89b6b60b4ca608d460b197e726d440a6/pkgs/build-support/trivial-builders/default.nix#L93-L94)).

Rule of thumb: if the buildscript runs a command that you wouldn't think twice about running locally, because it if fairly quick, then it is _probably_ better to use `runCommandLocal` to avoid waiting for network requests to binary substituters and remote builders.

Alternatively, if the derivation's inputs (e.g. source files) change so often that it is unlikely there will be a binary cache available, then it _may_ be worth considering `runCommandLocal`.

PR #408, opened 2026-05-28
don't derive Typeable

figsoda:
fixes nixfmt-static: https://github.com/NixOS/nixfmt/actions/runs/26479709024/job/77973918363

not sure why the static build behaves differently here, maybe it would be worth testing it in ci?

### DIFF
--- a/main.nix
+++ b/main.nix
@@ -73,7 +73,7 @@ let
   };
 
   checks = {
-    inherit build;
+    inherit build buildStatic;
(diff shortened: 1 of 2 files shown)


jfly:
Thanks for the PR!

> not sure why the static build behaves differently here, maybe it would be worth testing it in ci?

Yes, could you add this to prove it now builds?

figsoda:
added, hopefully enough stuff are cached that it finishes in a reasonable amount of time

jfly (approved):

PR #409, opened 2026-05-31
fix darwin build

figsoda:
also enables ci (only the `nix-build -A ci` part) on darwin, we can test aarch64-linux too but it might be out of scope for this pr

https://github.com/haskell/cabal/issues/4739#issuecomment-359209133

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -16,7 +16,18 @@ permissions:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    name: check ${{ matrix.system }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+          - system: aarch64-darwin
+            runner: macos-latest
+          - system: aarch64-linux
+            runner: ubuntu-24.04-arm
+          - system: x86_64-linux
+            runner: ubuntu-latest
+
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
         if: github.event_name != 'pull_request_target'
@@ -42,7 +53,7 @@ jobs:
           skipPush: ${{ github.event_name != 'push' }}
 
       - name: checks
-        run: nix-build -A ci
+        run: nix-build -A ci --argstr system "${{ matrix.system }}"
 
   nixpkgs-diff:
     runs-on: ubuntu-latest
--- a/nixfmt.cabal
+++ b/nixfmt.cabal
@@ -38,6 +38,12 @@ common warnings
     ghc-options:
       -Werror
 
+  -- the executable name nixfmt and the library namespace Nixfmt can collide on
+  -- darwin due to case insentivity on APFS
+  if os(darwin)
+    ghc-options:
+      -optP-Wno-nonportable-include-path
+
 executable nixfmt
   import: warnings
 


figsoda:
not sure if `buildStatic` would be useful on darwin, maybe we should remove it

nixfmt.cabal
```diff
@@ -38,6 +38,12 @@ common warnings
     ghc-options:
       -Werror
 
+  -- the executable name nixfmt and the library namespace Nixfmt can collide on
+  -- darwin due to case insentivity on APFS
```
MattSturgeon, line 42: Should we fix the underlying naming conflict?

.github/workflows/main.yml
```diff
@@ -16,7 +16,13 @@ permissions:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner:
+          - macos-latest
+          - ubuntu-latest
```
MattSturgeon: Could we do:

```suggestion
      matrix:
        include:
          - system: aarch64-darwin
            runner: macos-latest
          - system: x86_64-linux
            runner ubuntu-latest
```
And then use `name: check ${{ matrix.system }}` to set the job name too?

.github/workflows/main.yml
```diff
@@ -16,7 +16,13 @@ permissions:
 
 jobs:
   check:
-    runs-on: ubuntu-latest
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        runner:
+          - macos-latest
```
MattSturgeon: > we can test aarch64-linux too but it might be out of scope for this pr

Agreed, but I'd be happy to merge it as part of this PR if that's less friction than making a second pr.

MattSturgeon (commented):
Thanks for the quick fix!

> not sure if `buildStatic` would be useful on darwin, maybe we should remove it

So long as we get enough cache hits that it doesn't impact CI time-to-green I don't feel strongly.

If we ever plan to attach other platforms' static builds to releases, then CI should build them too. If not, I agree it's pointless work.

We could add a `build_attrs` field to the job matrix if that helps? Or just make the `ci` attr's content platform-specific.

nixfmt.cabal
figsoda: changing to `executable: Nixfmt` would fix the conflict, but we would need an additional step moving the binary from `Nixfmt` to `nixfmt`. I looked through https://github.com/haskell/cabal/issues/4739 and some relevant threads but couldn't find a better alternative

.github/workflows/main.yml
figsoda: done

.github/workflows/main.yml
figsoda: added aarch64-linux

figsoda:
> So long as we get enough cache hits that it doesn't impact CI time-to-green I don't feel strongly.

buildStatic is taking a long time on darwin-build-box due to the lack of existing cache, this might be an issue whenever nixpkgs is bumped

.github/workflows/main.yml
```diff
@@ -16,7 +16,18 @@ permissions:
...
-    runs-on: ubuntu-latest
+    name: check ${{ matrix.system }}
+    runs-on: ${{ matrix.runner }}
+    strategy:
+      matrix:
+        include:
+          - system: aarch64-darwin
+            runner: macos-latest
+          - system: aarch64-linux
+            runner: ubuntu-24.04-arm
```
jfly, line 27: There isn't a `ubuntu-larest-arm`? Bummer.

.github/workflows/main.yml
```diff
@@ -60,7 +71,7 @@ jobs:
         id: fc
         with:
           issue-number: ${{ github.event.pull_request.number }}
-          comment-author: 'github-actions[bot]'
+          comment-author: "github-actions[bot]"
```
figsoda: oops accidentally changed this with my formatter

nixfmt.cabal
jfly: What a mess. This seems like a fine solution.

jfly (approved):

.github/workflows/main.yml
MattSturgeon: Correct

.github/workflows/main.yml
```diff
@@ -42,7 +53,7 @@ jobs:
           skipPush: ${{ github.event_name != 'push' }}
 
       - name: checks
-        run: nix-build -A ci
+        run: nix-build -A ci --argstr system "${{ matrix.system }}"
```
MattSturgeon, line 56: This is totally fine here, as `matrix.system` is a controlled value.

But generally speaking, I try to avoid interpolating GHA expressions into shell scripts (any scripts, tbh).

Since these values don't contain _any_ shell syntax, we could even drop the quotes to avoid the illusion of safety.

The robust general-case solution would be to pass `matrix.system` in via `env`.

No changes needed, just commenting to raise awareness.

MattSturgeon (approved):
> > So long as we get enough cache hits that it doesn't impact CI time-to-green I don't feel strongly.
> 
> buildStatic is taking a long time on darwin-build-box due to the lack of existing cache, this might be an issue whenever nixpkgs is bumped

Let's try it out and play it by ear.

Thanks again!

PR #410, opened 2026-05-31
ci: skip static build on darwin 

figsoda:
static pretty-simple (dependency) fails to build on darwin: https://github.com/NixOS/nixfmt/actions/runs/26721856555/job/78750249867#step:6:47658

i can't think of a trivial fix, disabling it for now so 1.3.1 can be released

### DIFF
--- a/main.nix
+++ b/main.nix
@@ -138,6 +138,12 @@ let
     };
     treefmt = treefmtEval.config.build.check source;
   };
+
+  skippedChecks = lib.optionals pkgs.stdenv.hostPlatform.isDarwin [
+    # static pretty-simple fails to build on darwin
+    # https://github.com/NixOS/nixfmt/issues/411
+    "buildStatic"
+  ];
 in
 {
   packages = {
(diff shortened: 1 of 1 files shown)


main.nix
```diff
@@ -137,6 +137,10 @@ let
       installPhase = "touch $out";
     };
     treefmt = treefmtEval.config.build.check source;
+  }
+  # static pretty-simple fails to build on darwin
```
jfly: Could you file an issue with some details and link to it here?

figsoda: done #411

main.nix
MattSturgeon: I'd prefer we handle the ci-specific stuff here, in the `ci` attribute.

Something like:

```suggestion
  ci = pkgs.linkFarm "ci" ci;
```

And up in the let-in block:
```nix
ci = removeAttrs checks (
  if pkgs.stdenv.hostPlatform.isDarwin then
    [ "buildStatic" ]
  else
    [ ]
);
```
For example.


That way, locally, we could still _attempt_ to build checks not built by ci. Not too important here, as `buildStatic` is available through `packages`, but it sets a precedent.

figsoda: opted for a slightly more extendable version

MattSturgeon (approved):

PR #414, opened 2026-06-04
fix(readme): changed editor install instructions for nvim 0.11

lwctr:
neovim 0.11 changed the way of configuring lsp's to now use vim.lsp.config(), with this require("lspconfig") is now deprecated. (I do not know if the changes affects non-ls).

PR #415, opened 2026-06-05
docs: improve Neovim LSP instructions

MattSturgeon:
Follow up to #414, cc @lwctr

See https://neovim.io/doc/user/lsp/#vim.lsp.config()


jfly (approved):
Thank you!

PR #416, opened 2026-06-11
Drop npins & cleanup flake

MattSturgeon:
### Summary

Drop npins, in favor of using the flake lock.

### Pins
The actual pinned revisions are unchanged, but the Nixpkgs input has a different `narHash` when fetched from github.com instead of nixos.org.

### Dependabot
This will enable dependabot to update our inputs for us, as we already have it configured: https://github.com/NixOS/nixfmt/blob/1d1bf077b9a6e675e7558db04b4c553e21858253/.github/dependabot.yml#L3-L6

### Non-flake usage
The non-flake codepath now uses flake-compat to fetch inputs.

We only use flake-compat to evaluate **inputs, not outputs**; using flake-compat for outputs would a) make overriding `nixpkgs` more difficult and b) make building on unsupported `system`s more difficult.

### Reviewing
The patches should be fairly understandable either reviewed commit-by-commit or as a single unified diff.

---

### Nixpkgs CI

cc @mdaniels5757 @philiptaron I'm thinking of a similar _"use flake-compat for inputs" and "use dependabot for updates"_ pattern in nixpkgs-ci. What do you think of the idea?

For nixpkgs-ci it may make sense to read the lockfile directly, without flake-compat:
```nix
  lock = builtins.fromJSON (builtins.readFile ./flake.lock); 
  inputNodes = builtins.mapAttrs (_: nodeName: lock.nodes.${nodeName}) lock.nodes.${lock.root}.inputs;
  inputs = builtins.mapAttrs (_: node: builtins.fetchTarball {/*...*/) inputNodes;
```

If we can assume we don't need the `follows` system, don't need to evaluate inputs' respective flake outputs, and all inputs are from GitHub.


jfly (approved):
Thanks!

> feat(shell): add inputs parameters to `shell.nix`

I don't understand the purpose of this change. Could you explain it? Ideally in the commit message.

MattSturgeon:
> > `` feat(shell): add inputs parameters to `shell.nix` ``
> 
> I don't understand the purpose of this change. Could you explain it? Ideally in the commit message.

Mostly for parity with `default.nix`, where `nixpkgs` can be explicitly overridden by users.

I'm unsure whether there's a clear use-case, happy to drop that commit if you prefer.

jfly:
> I'm unsure whether there's a clear use-case, happy to drop that commit if you prefer.

I say drop it. If you want to keep it for symmetry, just explain that's why in the commit.

philiptaron:
> nixpkgs-ci. What do you think of the idea?

Assuming you mean `nixpkgs-vet`, quite into it.

MattSturgeon:
> Assuming you mean `nixpkgs-vet`, quite into it.

I actually meant replacing `ci/pinned.json` with `ci/flake.lock` in Nixpkgs.[^1]

But, sure, nixpkgs-vet is also a candidate: https://github.com/NixOS/nixpkgs-vet/pull/257

[^1]: Without introducing a hard requirement to _consume_ CI via flakes, as-per the impl in this PR.

PR #423, opened 2026-07-01
fix(sync-pr): build each Nixfmt revision via pure flakes

MattSturgeon:
Me and @jfly worked on this together over jitsi.

We can gain some purity guarantees by building the PR's Nixfmt revisions using a flakes-based workflow. Tried to keep other changes to a minimum.

We did some smoke testing on my fork, at https://github.com/MattSturgeon/nixfmt/pull/1
Not thoroughly tested, but hopefully the changes are straightforward enough.


jfly (approved):

infinisil:
@jfly informed me that it would be good to rotate the infinixbot token after this. 

Just did that :)

PR #430, opened 2026-07-11
Tab support

toastal:
Adds indentation character of `\t` or ` `. When `--tabs` is passed, the indentation width is set to 1 for 1 tab. The goal is accessibility as tabs can be adjusted downstream by users to suit their needs (Braille user may only need 1 width, low vision may need 8, & I find myself
needing a value beyond 2).

Many programming / scripting languages — including Nix & Bash used in stdenv — support both tabs & spaces. Bugs in Nix’s indented strings still exist with tabs, but files using only tabs work just fine so long as the users isn’t mixing tabs / spaces. However, since so many
languages support both, this largely is a non-issue.

As for `width`, I think it’s just fine to count 1 tab as 1 character since it is configurable & this approach is simplest. It can be fought over later as this isn’t the default setting anyhow, but gives users accessibility *now*. See: https://alexandersandberg.com/articles/default-to-tabs-instead-of-spaces-for-an-accessible-first-environment/.

I have been using this patchset on my own projects for a couple of weeks.

[^1]

[^1]: Please consider [giving up MS GitHub](https://sfconservancy.org/GiveUpGitHub/) or offering a non-proprietary, non-US-corporate-controlled mirror for this free software project. I wish to delete this Microsoft account in the future, but I need more projects like this to support alternative methods to send patches & contribute.

### DIFF
--- a/src/Nixfmt/Predoc.hs
+++ b/src/Nixfmt/Predoc.hs
@@ -52,8 +52,11 @@ import Data.Text as Text (Text, concat, length, replicate, singleton, strip, str
 import GHC.Stack (HasCallStack)
 import Nixfmt.Types (
   Ann (..),
+  Indentation (..),
   LanguageElement,
   Trivium (..),
+  indentChar,
+  indentWidth,
   mapAllTokens,
   mapFirstToken,
   removeLineInfo,
@@ -418,8 +421,8 @@ mergeSpacings Hardspace (Newlines x) = Newlines x
 mergeSpacings _ (Newlines x) = Newlines (x + 1)
 mergeSpacings _ y = y
 
-layout :: (Pretty a, LanguageElement a) => Int -> Int -> Char -> Bool -> a -> Text
-layout width indentWidth indentChar strict a = finalize $ layoutGreedy width indentWidth indentChar doc
+layout :: (Pretty a, LanguageElement a) => Int -> Indentation -> Bool -> a -> Text
+layout width indentation strict a = finalize $ layoutGreedy width indentation doc
   where
     doc =
       fixup
@@ -580,10 +583,13 @@ indent c n = Text.replicate n (Text.singleton c)
 type St = (Int, NonEmpty (Int, Int), Bool)
 
 -- tw   Target Width
-layoutGreedy :: Int -> Int -> Char -> Doc -> Text
-layoutGreedy tw iw ic doc =
+layoutGreedy :: Int -> Indentation -> Doc -> Text
+layoutGreedy tw indentation doc =
   Text.concat $ evalState (go [Group RegularG doc] []) (0, NonEmpty.singleton (0, 0), False)
   where
+    iw = indentWidth indentation
+    ic = indentChar indentation
+
     -- Simple helpers around `put` with a tuple state
     -- NOTE: making putL strict removes the allocation for the Int
     putL v = modify' (\(_, indents, suppress) -> (v, indents, suppress))
--- a/src/Nixfmt/Predoc.hs
+++ b/src/Nixfmt/Predoc.hs
@@ -44,11 +44,11 @@ import Data.Function ((&))
 import Data.Functor (($>), (<&>))
 import Data.Functor.Identity (runIdentity)
 import Data.List (intersperse)
-import Data.List.NonEmpty (NonEmpty (..), singleton, (<|))
+import Data.List.NonEmpty (NonEmpty (..), (<|))
 import Data.List.NonEmpty qualified as NonEmpty
 import Data.Maybe (fromMaybe)
 import Data.Sequence qualified as Seq
-import Data.Text as Text (Text, concat, length, replicate, strip, stripStart, takeWhileEnd)
+import Data.Text as Text (Text, concat, length, replicate, singleton, strip, stripStart, takeWhileEnd)
 import GHC.Stack (HasCallStack)
 import Nixfmt.Types (
   Ann (..),
@@ -418,8 +418,8 @@ mergeSpacings Hardspace (Newlines x) = Newlines x
 mergeSpacings _ (Newlines x) = Newlines (x + 1)
 mergeSpacings _ y = y
 
-layout :: (Pretty a, LanguageElement a) => Int -> Int -> Bool -> a -> Text
-layout width indentWidth strict a = finalize $ layoutGreedy width indentWidth doc
+layout :: (Pretty a, LanguageElement a) => Int -> Int -> Char -> Bool -> a -> Text
+layout width indentWidth indentChar strict a = finalize $ layoutGreedy width indentWidth indentChar doc
   where
     doc =
       fixup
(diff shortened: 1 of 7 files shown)


src/Nixfmt/Predoc.hs
```diff
@@ -47,11 +47,11 @@ import Data.Function ((&))
 import Data.Functor (($>), (<&>))
 import Data.Functor.Identity (runIdentity)
 import Data.List (intersperse)
-import Data.List.NonEmpty (NonEmpty (..), singleton, (<|))
+import Data.List.NonEmpty (NonEmpty (..), (<|))
```
toastal, line 47: Drop as the compiler was throwing an ambiguity error.

dyegoaurelio:
Thanks for your contribution!

However, I don't think we can merge it because it's not advisable to format nix code with tabs.

See https://github.com/NixOS/nixfmt/issues/77#issuecomment-3177347091 , https://github.com/NixOS/nix/issues/7834 https://github.com/NixOS/nix/issues/3759

toastal:
1. Lix fixed the issue (not all `nixfmt` users use Nix C++)
2. If not mixing tabs / spaces this doesn’t become a problem as at some point *some* of the older issues are addressed
3. As a non-default setting, users can opt into it if they know the rules (see JavaScript’s ASI where you should know the rules/gotchas regardless)—so there could be some updated option text to warn users
4. Then at least this patch/fork will exist for those that wish to use tabs in their projects like I have been doing

toastal:
For reference, Alejandra supports tabs as an option: https://github.com/kamadorueda/alejandra/blob/8c4a4a572bee519b04e9bb9207e7d993f55ecb4f/src/alejandra/src/config.rs#L18-L26

piegamesde:
> Lix fixed the issue

Tabs in Nix lang indented strings are fundamentally broken, and there is **no** way to fix this without introducing either a new string literal syntax or a breaking change. Lix has not fixed this, since it can't without risking accidentally breaking existing code (ignoring experimental feature flags), and all it has done so far is to disarm the worst footguns by turning them into an error.

> If not mixing tabs / spaces this doesn’t become a problem as at some point some of the older issues are addressed

This is news to me, but I might be forgetting or confusing things so feel free to elaborate

Relevant Lix code for reference: https://git.lix.systems/lix-project/lix/src/branch/main/lix/libexpr/parser/grammar.hh#L265, https://git.lix.systems/lix-project/lix/src/branch/main/lix/libexpr/parser/state.hh#L480

-----

Wanting to support tabs indentation may be worth pursuing nevertheless, but it is of utmost importance to inform users about the true gotchas, which include but are not limited to "changing a level of indentation may cause derivations to change and rebuild"

jfly (commented):
Skimmed the code, looks reasonable to me.

High level, I'm supportive of this feature, so long as it's off by default and hidden behind a flag (as you've done). Could you update the `--help` text to explain the consequences of using this feature? Are there nix programs that would evaluate differently if formatted with `nixfmt --tabs`? At the very least we document the possibility. Ideally we'd error out if given such an input program.

toastal:
There is probably a better invariant modeling too

```ocaml
type indentation =
	| Spaces of int
	| Tabs
```

But my Haskell is very rusty (definitely no pun intended) & this is effectively an MVP.

piegamesde:
> Are there nix programs that would evaluate differently if formatted with nixfmt --tabs?

Yes, that's what I meant with "it may change derivations".

```nix
''
 foo
''
```

is `"foo"`, but with tabs it would be `"\tfoo"`. The only reason why it doesn't matter for most practical purposes is that indented strings are primarily used for generating code in other programming languages, which happen to be mostly agnostic over tabs vs. spaces.

jfly:
Thanks for the concrete example, @piegamesde. Sorry for not reading all the linked discussions.

I sure would prefer an error rather than changing primitive values in the input program. @toastal, WDYT?

toastal:
I pushed an attempt at encoding the invariant. Seems `cmdargs` doesn’t allow any mutual exclusivity? Weird. If the commit is mostly noise & not helpful, as it is separate, it can be dropped.

toastal:
I added a `WARNING` but I don’t know or think the text is good enough.

jfly:
For the record, @MattSturgeon and I discussed this very briefly today (lost most of our time discussing other PRs, sorry). We both are unwilling to ship a formatter that changes the actual byte content of string literals in input programs. I just checked, and the code in this PR does exactly that. For example, here's a given input program:

`tabs-in.nix`:

```nix
let str = ''
  foo
'';
in builtins.toJSON str
```

The string literal in the original program (`str`) is a 4 character string starting in `f` and ending with a newline:

```console
$ cat tabs-in.nix | nix eval --file - --raw
"foo\n"
```

When formatted with `nixfmt`, it remains that exact same 4 character string:

```console
$ cat tabs-in.nix | ./result/bin/nixfmt - | nix eval --file - --raw
"foo\n"
```

When formatted with `nixfmt --tabs`, it becomes a 7 character string starting and ending with tabs:

```console
$ cat tabs-in.nix | ./result/bin/nixfmt - --tabs | nix eval --file - --raw
"\t\tfoo\n\t"
```

(Tested with `nix` 2.34.8. I get the exact same behavior with `lix` 2.95.2.)

This is not OK for our formatter to do. I see 2, *maybe* 3 paths forward:

1. Do not land this PR.
2. Tweak this PR to preserve the exact byte-for-byte contents of string literals. The simplest option would be to not change their indentation at all, but for completeness sake, I suppose there are alternatives that involve spaces.
3. Ship this PR, but rename `--tabs` to something obnoxious like `--dangerously-potentially-change-the-contents-of-my-string-literals-by-indenting-with-tabs`.

We'll try to discuss this in our next team meeting.

toastal:
I caught this this weekend when trying to mess with NixOS tests where the injected Python prefix snippet messed created an issue. Many languages (such as Bash) aren’t indentation sensitive enough for it to matter.

This issue isn’t nixfmt’s concern, but this should be a bug in upstreams, no?

MattSturgeon:
> this should be a bug in upstreams, no?

https://github.com/NixOS/nix/issues/7834

toastal:
I think that is the issue. I always seem to have trouble finding it…

jfly:
> This issue isn’t nixfmt’s concern, but this should be a bug in upstreams, no?

I maintain that our formatter must preserve the behavior of the input program. If https://github.com/NixOS/nix/issues/7834 were a recent regression, we could debate which *version* of nix we use to make that determination (and if we are interested in accommodating nix-like interpreters, such as lix), but AFAIK that's just not the case here. (I don't actually understand what Lix has done about this. I understood they've made some change, but I didn't observe a difference in [the tests I did above](https://github.com/NixOS/nixfmt/pull/430#issuecomment-5038531297).)

I do personally hope that the situation with https://github.com/NixOS/nix/issues/7834 evolves. I believe Nix should have a mechanism for changing bad decisions from the past, rather than being trapped by old decisions in the name of reproducibility. Ideally something more like Rust's epochs and less like Python's 2 -> 3.

MattSturgeon:
> This is not OK for our formatter to do. I see 2, _maybe_ 3 paths forward:
>
> 1. Do not land this PR.
> 2. Tweak this PR to preserve the exact byte-for-byte contents of string literals. The simplest option would be to not change their indentation at all, but for completeness sake, I suppose there are alternatives that involve spaces.
> 3. Ship this PR, but rename `--tabs` to something obnoxious like `--dangerously-potentially-change-the-contents-of-my-string-literals-by-indenting-with-tabs`.

Discussed in today's formatting team meeting. A month on we haven't seen any discussion towards points 2 or 3, so we're going to go with option 1 and close the PR for now.

PR #435, opened 2026-07-17
fix: exponential parse time for nested set parameter defaults

### LINKED ISSUE #434: nixfmt hangs on deeply nested set-pattern lambda defaults
This issue was reported by project glasswing.

------
#### `TLDR.md`

In `abstraction` at Parser.hs:455, `contextParameter` wraps `setParameter` in a `try` that backtracks when no `@` follows, and inside `setParameter` the `commaAttrs`/`lastAttr` split also backtracks and re-parses each `name ? default`. Combined, every default expression in a `{a ? X}: ...` lambda is parsed four times.

##### Mitigation

- nixfmt is an offline developer/CI tool, not a network-exposed service
- Requires the victim to voluntarily format an attacker-supplied file
- Effect is CPU hang / DoS only — no memory corruption, code execution, or data disclosure
- Typically run in contexts (editor subprocess, CI step) that can be cancelled or will time out

##### ANT-2026-SR6N5S8A: Exponential parser backtracking on nested lambda parameter defaults

In `abstraction` at Parser.hs:455, `contextParameter` wraps `setParameter` in a `try` that backtracks when no `@` follows, and inside `setParameter` the `commaAttrs`/`lastAttr` split also backtracks and re-parses each `name ? default`. Combined, every default expression in a `{a ? X}: ...` lambda is parsed four times. When the default is itself another set-pattern lambda, this multiplies through `attrParameter → expression → abstraction`, giving 4^d parses of the innermost default at depth d. A ~150-byte valid Nix file nested ~25 levels deep triggers ~10^15 parse invocations, causing nixfmt to spin on CPU indefinitely with no error. Megaparsec has no memoization to cap the re-parsing.

###### Target

**Project:** NixOS/nixfmt  
**Location:** `src/Nixfmt/Parser.hs:448`  

###### Technical Details

The root cause is redundant backtracking at two layers: `try (ContextParameter <$> setParameter <*> symbol TAt <*> idParameter)` (line 448) fully parses the set-pattern then discards it when `@` is absent, and `many $ try $ attrParameter (Just $ symbol TComma)` (line 441) parses `name ? default` then backtracks on the missing trailing comma before `lastAttr` re-parses it. Each layer doubles the work on the default expression (4× total), and because defaults can recursively contain set-pattern lambdas, the cost is exponential in nesting depth.

###### Reproduction

1. Craft a syntactically-valid Nix file with ~25 nested set-pattern lambda defaults, e.g. `{a?{a?...{a?1}:1...}:1}:1` (~150 bytes).
2. Deliver the file to the victim (commit to a public repo, PR, shared project).
3. Victim's editor/LSP format-on-save, git pre-commit hook, or CI treefmt step runs `nixfmt` on the file.
4. nixfmt enters ~4^25 ≈ 10^15 parse invocations and consumes 100% CPU effectively forever until killed.

See the attached PoC.

###### Suggested Fix

Restructure the parameter grammar so a set-pattern and each default expression are parsed at most once per position: parse the set-pattern once and decide the optional `@ident` suffix by lookahead instead of `try`-wrapping the whole thing, and parse `name ? default` once with an optional trailing comma rather than backtracking on comma absence.

###### Acknowledgement

This vulnerability was discovered by Claude, Anthropic's AI assistant, and triaged by the Anthropic security team in collaboration with Anthropic Research. Please direct questions to security-cvd@anthropic.com and reference ANT-2026-SR6N5S8A.

---

**Reference:** ANT-2026-SR6N5S8A  
**Anthropic CVD Policy:** https://anthropic.com/security/cvd-policy

dyegoaurelio:
Parsing a set parameter whose attribute defaults contain further set parameters took time exponential in the nesting depth. Such patterns now parse in linear time.

Two backtracking points each threw away and re-parsed an unbounded amount of successfully parsed input, multiplying into 4^depth:

- A context parameter (`{ ... }@x`) was `try`ed as a whole, so for a plain `{ ... }` the entire set parameter was parsed just to discover there is no `@` behind it, discarded, and parsed again from scratch.
- Each attribute was `try`ed together with its trailing comma, so the comma-less last attribute of every set parameter was parsed twice, including its arbitrarily large default expression.

The parameter parsers are restructured to commit early instead of backtracking: a set parameter is parsed exactly once and a following `@` then promotes it to a context parameter; an attribute list commits to each attribute before looking for its comma. In both spots the decision after the committed prefix hinges on a single token that fails without consuming input, and the one remaining `try` backtracks at most one identifier (`x:` vs `x@{ ... }:`).

The new test case would take years to parse before this changes.

closes #434


piegamesde (commented):
The fix is technically sound, but I consider it a big downgrade in code readability and maintainability. I think having a human fix the bug would lead to a much better outcome than this

For context, the original parser here is quite naive (hence the performance bug), and I consider this fix an equally quite naive transformation on the original code. The result is difficult to grasp (not that it was easy in the first place with Haskell parsing code). Taking a step back and re-thinking the parser for this syntax from scratch should lead to something much cleaner, I'd take the Lix grammar as guide for that

dyegoaurelio:
The issue was reported by project glasswing, but this fix wasn't generated by an agent.

**I do use ai tools, but I did write this patch by myself.**

I'm not very fluent with haskell (and with writing parsers in general) and I do think the nixfmt codebase is pretty good, so I don't usually lean  towards refactoring much stuff.

I honestly don't consider it a big readability downgrade, I did try to make the code as straight forward as possible.

The existing code had a subparser for each param type, since the fix was to stop backtracking on set/ctx params, I defined a single parser for set and context params which upgrades a set to a ctx param if it finds an `@`.

I think this behavior is pretty clear now.
```hs
parameter :: Parser Parameter
parameter =
  try ctxOrSetParameter
    <|> idParameter
```

I removed the subparsers from the global scope because having `ctxOrSetParameter` there would be pretty asymmetrical with the other globally defined parsers

Nixfmt is my first experience working with haskell and working with parsers/formatters, so I respect the input from someone with more expertise in the domain and  would appreciate if you could provide a more comprehensive review for this pr.

jfly (approved):
@MattSturgeon and I scratched our heads over this for ~1 hour. Not a comment on the quality of the code, just a comment on our collective lack of haskell knowledge.

This code certainly seems to have the right "shape" for the solution. Just tracking where the `try`s have moved smells right.

Even if there's a much cleaner solution behind a big refactor of the project, we agree it's better to land this PR as is. Doesn't preclude a refactor in the future.

As I am fond of saying, our test suite is our most valuable asset in this repo, so this will help any hypothetical future refactors.

MattSturgeon (approved):
SGTM! Let's give it a go.

piegamesde:
Sorry for dropping you here, I'll try to rework this section when I get to it soon-ish

PR #440, opened 2026-07-22
Rework function attribute parsing

piegamesde:
Follow-up to #435, CC @dyegoaurelio

Turns out this one was way more icky than anticipated, and also I've gotten drawn into doing related refactorings while I'm at it. In retrospect I could probably do this change with a more minimal diff, but oh well. The remaining syntax change in the output is mostly a TODO, in the sense that I don't know how we want to handle those cases (see the newly added tests). Right now I'm erring on the side of "doesn't matter nobody would put comments in here anyways so let's keep the code simple", but if people have opinions I can still polish that

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -408,21 +408,26 @@ containsDirective = any $ \case
   _ -> False
 
 instance Pretty Parameter where
-  -- param:
+  -- Simple: `param`
   pretty (IDParameter i) = pretty i
-  -- {}:
-  pretty (SetParameter bopen [] bclose) =
+
+  -- Empty attrs: `{}`
+  pretty (SetParameter Nothing bopen [] bclose Nothing) =
     group $ pretty (moveTrailingCommentUp bopen) <> sep <> pretty bclose
     where
       -- If the braces are on different lines, keep them like that
       sep = if sourceLine bopen /= sourceLine bclose then hardline else hardspace
-  -- { stuff }:
-  pretty (SetParameter bopen attrs bclose) =
-    group $
-      pretty (moveTrailingCommentUp bopen)
-        <> surroundWith sep (nest $ sepBy sep $ handleTrailingComma $ map moveParamAttrComment $ moveParamsComments attrs)
-        <> nest (pretty (preTrivia bclose))
-        <> pretty (bclose{preTrivia = []})
+
+  -- Non-empty attrs: `{ stuff }`, `{ stuff }@param`, `param@{ stuff }`
+  pretty (SetParameter paramLeft bopen attrs bclose paramRight) =
+    pretty paramLeft
+      <> group
+        ( pretty (moveTrailingCommentUp bopen)
+            <> surroundWith sep (nest $ sepBy sep $ handleTrailingComma $ map moveParamAttrComment $ moveParamsComments attrs)
+            <> nest (pretty (preTrivia bclose))
+            <> pretty (bclose{preTrivia = []})
+        )
+      <> pretty paramRight
     where
       -- pretty all ParamAttrs, but mark the trailing comma of the last element specially
       -- This is so that the trailing comma will only be printed in the expanded form
@@ -437,8 +442,6 @@ instance Pretty Parameter where
         if canFlattenAttrs bopen attrs bclose
           then line
           else hardline
-  pretty (ContextParameter param1 at param2) =
-    pretty param1 <> pretty at <> pretty param2
 
 -- Function application
 -- Some example mapping of Nix code to Doc (using brackets as groups, but omitting the outermost group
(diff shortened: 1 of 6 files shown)


piegamesde:
Can we please have auto-format in pre-commit hook?

test/diff/pattern/out-pure.nix
```diff
@@ -797,4 +797,178 @@
...
+  )
+  ({ }@foo: bar)
+  (
+    # comment
+    { }@foo: bar
+  )
+  (
+    {
+    } # comment
+    @foo:
```
MattSturgeon, line 961: How come this does a hard-break inside the set-pattern parameter? I'd intuitively expect:
```suggestion
    { } # comment
    @foo:
```

src/Nixfmt/Types.hs
```diff
@@ -222,31 +222,37 @@ data Term
...
 data Parameter
-  = IDParameter Leaf
-  | SetParameter Leaf [ParamAttr] Leaf
-  | ContextParameter Parameter Leaf Parameter
+  = -- Also called "simple pattern" in Nix
+    IDParameter Leaf
+    -- Also called "attrs pattern" in Nix
+    -- Can be either `{ ... }`, `{ ... }@ident`, `ident@{ ... }`
+    -- Cannot be `ident@{}@ident`, though our CST could theoretically represent it
+  | SetParameter (Maybe (Leaf, Leaf)) Leaf [ParamAttr] Leaf (Maybe (Leaf, Leaf))
```
MattSturgeon, line 239: I like how `SetParameter` now represents all forms of set-pattern parameters, rather than having to reason about `ContextParameter` separately.

test/diff/pattern/out-pure.nix
dyegoaurelio: fwiw, I ran this patch's `in.nix` against master and this was the only divergence

```patch
diff --git a/test/diff/pattern/out-pure.nix b/test/diff/pattern/out-pure.nix
index 9161d28..464b0aa 100644
--- a/test/diff/pattern/out-pure.nix
+++ b/test/diff/pattern/out-pure.nix
@@ -956,7 +956,8 @@
     { }@foo: bar
   )
   (
-    { } # comment
+    {
+    } # comment
     @foo:
     bar
   )
diff --git a/test/diff/pattern/out.nix b/test/diff/pattern/out.nix
index 9161d28..464b0aa 100644
--- a/test/diff/pattern/out.nix
+++ b/test/diff/pattern/out.nix
@@ -956,7 +956,8 @@
     { }@foo: bar
   )
   (
-    { } # comment
+    {
+    } # comment
     @foo:
     bar
   )

```

src/Nixfmt/Pretty.hs
```diff
@@ -666,7 +669,8 @@ isAbsorbableExpr expr = case expr of
   (Abstraction (IDParameter _) _ (Term t)) | isAbsorbable t -> True
   (Abstraction (IDParameter _) _ body@(Abstraction{})) -> isAbsorbableExpr body
   -- Absorb function declarations with attribute set parameters if the attribute set is simple enough, and the body is absorbable.
-  (Abstraction (SetParameter bopen attrs bclose) colon body)
+  -- Parameters with a context identifier are never absorbed.
+  (Abstraction (SetParameter Nothing bopen attrs bclose Nothing) colon body)
```
dyegoaurelio, line 673: I think it's better to move this logic to `canFlattenAttrs`


```suggestion
  (Abstraction (SetParameter paramLeft bopen attrs bclose paramRight) colon body)
    | canFlattenAttrs paramLeft bopen attrs bclose paramRight,
```

src/Nixfmt/Types.hs
```diff
@@ -222,31 +222,37 @@ data Term
...
+-- Also called "pattern" in Nix
 data Parameter
-  = IDParameter Leaf
-  | SetParameter Leaf [ParamAttr] Leaf
-  | ContextParameter Parameter Leaf Parameter
+  = -- Also called "simple pattern" in Nix
+    IDParameter Leaf
+    -- Also called "attrs pattern" in Nix
+    -- Can be either `{ ... }`, `{ ... }@ident`, `ident@{ ... }`
+    -- Cannot be `ident@{}@ident`, though our CST could theoretically represent it
```
dyegoaurelio, line 238: we could add a test asserting that `ident@{}@ident` gets rejected

dyegoaurelio (commented):
Thanks! Looks good to me.

I think that we can merge it once we fix https://github.com/NixOS/nixfmt/pull/440/changes#r3633975855

and maybe add a test case for `ident@{}@ident`

src/Nixfmt/Pretty.hs
dyegoaurelio: actually, this new signature would make `canFlattenAttrs` weird to use on the Pretty instance for Parameter above.

dyegoaurelio (approved):

MattSturgeon (approved):
Thanks!

PR #443, opened 2026-07-29
test: add failing idempotency case for trailing comment after multili…

Mic92:
…ne string

A line comment trailing the closing quote of a multiline "..." string becomes non-idempotent when reformatting moves the next item to the comment's column.
The second pass reattaches the comment as leading.

    [ "
    " # c
          t2
    ]

### DIFF
--- a/test/diff/comment/in.nix
+++ b/test/diff/comment/in.nix
@@ -165,4 +165,9 @@
   "
 " # c
   t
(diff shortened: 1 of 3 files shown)


jfly:
Confirmed this issue is still present after rebasing on top of `master`.

Issue #312, opened 2025-06-09
nixfmt converts symbolic links to a copy


badcold:
I've tried to reformat my machine's configuration.nix, which was a symbolic link and it made a copy of the file at the link's location.

```
[badcold@i7-3770k:~/demo]$ nixfmt --version
nixfmt nixpkgs-unstable-2025-04-04

[badcold@i7-3770k:~/demo]$ ls -la
total 16
drwxr-xr-x  4 badcold users 4096 Jun  9 11:11 .
drwx------ 72 badcold users 4096 Jun  9 11:10 ..
drwxr-xr-x  7 badcold users 4096 Jun  9 11:11 .git
drwxr-xr-x  2 badcold users 4096 Jun  9 11:10 elsewhere
lrwxrwxrwx  1 badcold users   18 Jun  9 11:11 test.nix -> elsewhere/test.nix

[badcold@i7-3770k:~/demo]$ nixfmt test.nix

[badcold@i7-3770k:~/demo]$ ls -la
total 20
drwxr-xr-x  4 badcold users 4096 Jun  9 11:12 .
drwx------ 72 badcold users 4096 Jun  9 11:10 ..
drwxr-xr-x  7 badcold users 4096 Jun  9 11:11 .git
drwxr-xr-x  2 badcold users 4096 Jun  9 11:10 elsewhere
-rw-r--r--  1 badcold users  120 Jun  9 11:12 test.nix
```

jfly:
+1 to what @MattSturgeon said about using treefmt. However, we would accept a PR that follows the symlink and formats the file it points at.

PR #455, opened 2026-09-01
[#312] Add --follow-symlinks flag

Sereja313:
Formatting a symlink replaces it with a regular file. That stays the default and is now warned about, `--follow-symlinks` formats the file the link points at instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -288,6 +288,8 @@ to return back to the unmerged state.
 
 * `nixfmt < input.nix` – reads Nix code from `stdin`, formats it, and outputs to `stdout`
 * `nixfmt file.nix` – format the file in place
+* `nixfmt --follow-symlinks link.nix` – if `link.nix` is a symbolic link, format the regular file it points at (resolving any symlinks along the way).
+  Without `--follow-symlinks`, the symlink is replaced by a regular file holding the formatted result, and a warning is printed to stderr
 
 ## Disabling Formatting
 
--- a/main/Main.hs
+++ b/main/Main.hs
@@ -3,7 +3,7 @@
 
 module Main where
 
-import Control.Monad (forM, unless)
+import Control.Monad (forM, unless, when)
 import Control.Monad.Trans.Class (lift)
 import Control.Monad.Trans.Except (ExceptT (ExceptT), runExceptT, throwE)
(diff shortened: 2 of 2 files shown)


dyegoaurelio (approved):

README.md
```diff
@@ -288,6 +288,7 @@ to return back to the unmerged state.
 
 * `nixfmt < input.nix` – reads Nix code from `stdin`, formats it, and outputs to `stdout`
 * `nixfmt file.nix` – format the file in place
+* `nixfmt --follow-symlinks link.nix` – format the file a symbolic link points at. Without the flag the link is replaced by a regular file holding the formatted result, and nixfmt warns on stderr that it did so
```
jfly: nit: The prose here is a little ambiguous: what if the file is a regular file? What if it's a symlink to another symlink?

Is the previous behavior deprecated? (Doesn't have to happen here, but I think it would be reasonable to deprecate it.)


```suggestion
* `nixfmt --follow-symlinks link.nix` – if `link.nix` is a symbolic link, format the regular file it points at (following any symlinks discovered along the way). Without the flag the link is replaced by a regular file holding the formatted result, and nixfmt warns on stderr that it did so
```

jfly (approved):
A couple comments for you, but LGTM. Thanks!

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Revision history for nixfmt
 
+## Unreleased
+
+- Added `--follow-symlinks`, which formats the file a symbolic link points at instead of replacing the link with a regular file. Without the flag the previous behaviour is kept, but nixfmt now warns when it replaces a symlink: <https://github.com/NixOS/nixfmt/issues/312>
```
MattSturgeon: Now we're using Knope, this needs moving to _either_ conventional commits or a `.changeset` file

MattSturgeon: I pushed up a conventional commit reword, so now the changelog will render as:

```markdown
### Features

- **support chained presence checks (a ? b ? c)** by Dyego Aurélio (bf52dec)
- **introduce `--follow-symlinks` option** by Sergey Gulin (608248d)
```

MattSturgeon (approved):
I iterated on @jfly's README suggestion and replaced the CHANGELOG entry with a conventional commit (that Knope will pick up on).

PR #456, opened 2026-09-01
fix: indent arguments of parenthesized function application

### LINKED ISSUE #385: Paren-wrapped function args get aligned with function
### Description

I've found formatting surprising if I have a multiline function expression that I pass arguments too (and the whole expression is also wrapped in parens). It's a mouthful. Look at the examples.

### Small example input

```nix
(
  (func-builder {
    bar = "baz";
  }) arg1 arg2
)
```

### Expected output

```nix
(
  (func-builder {
    bar = "baz";
  })
    arg1
    arg2
)
```

### Actual output

```nix
(
  (func-builder {
    bar = "baz";
  })
  arg1
  arg2
)
```

### Additional context

When not enclosed in parentheses, (e.g. top-level file expression), it is formatted as:
```nix
(func-builder {
  bar = "baz";
})
  arg1
  arg2
```

ashwinmathi:
Previously, `prettyApp` nested the function head as well as its arguments, so both sat at the same level. I believe this was related to fda8afa, which introduced `indentFunction` when nesting was column-based (`nest 2`, measured from a `base` anchor). eb732b1 later made `nest` ordinal and dropped `base`, leaving the head's nest as a duplicate of its caller's.

Both `indentFunction=True` callers already wrap the application in `nest`, so the head's own nest was redundant. Removing it nests arguments one level deeper.

Fixes #385

### DIFF
--- a/src/Nixfmt/Pretty.hs
+++ b/src/Nixfmt/Pretty.hs
@@ -476,7 +476,7 @@ prettyApp indentFunction pre hasPost f a =
       absorbApp (Application f' a') = group' Transparent (absorbApp f') <> line <> nest (group' Priority $ absorbInner a')
       -- First argument
       absorbApp expr
-        | indentFunction && null comment' = nest $ group' RegularG $ line' <> pretty expr
+        | indentFunction && null comment' = group' RegularG $ line' <> pretty expr
         | otherwise = pretty expr
 
       -- Render the inner arguments of a function call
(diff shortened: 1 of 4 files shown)


Sereja313 (approved):

dyegoaurelio (approved):

piegamesde (dismissed):
Impressive catch, especially given how unintuitive the code is here!

One nit though on the commit structure: The Nixfmt tests are Characterization tests, not unit tests. Therefore the flow is not "add failing test case, then the fix", but "add test case with wrong result, and the fix commit will contain the correction"

jfly (approved):
Thanks, @ashwinmathi!

PR #457, opened 2026-09-02
ci: publish using Knope

MattSturgeon:
Migrate the changelog and release process to [Knope](https://knope.tech).

This is based on what @mdaniels5757 implemented in [nixpkgs-vet](https://github.com/NixOS/nixpkgs-vet), and what I recently added to [Freecam](https://github.com/MinecraftFreecam/Freecam).

### Architecture

The `release.yml` workflow runs on all pushes.

Its first job, `version`, checks out `nixfmt.cabal` & `knope.toml`:
- Reads `base` from `knope.toml`'s `CreatePullRequest` step to determine if this is a release branch.
- Reads `version:` from `nixfmt.cabal`, similar to the existing workflow.
- Queries GitHub's `getReleaseByTag` to determine whether the current version has already been published to GitHub Releases.
  - Draft releases are ignored, because it only has `contents: read` permissions.

This gives us three release states:
- **Not a release branch:** if `knope.toml`'s `base` is not `GITHUB_REF_NAME`, then this is a feature branch not a release branch.
- **Current version is already published:** prepare the next release by creating or updating the release-preview PR.
- **Current version is unpublished:** publish the release that has been merged to `master`.

#### Create release PR

The `Create release PR` job keeps a release-preview PR up to date, so there should normally always be an open PR showing what the next release will contain.

After checkout and setup, it runs `knope prepare-release` to generate the next version and changelog, then commits those changes as `chore: release 1.4.0` and pushes them to the `prepare-release` branch before creating or updating the PR.

#### Publish release

The `Publish release` job actually publishes a release whose version bump has been merged to `master`. This typically happens after a release PR is merged.

After checkout and setup, it builds the static `nixfmt` artifact and then runs `knope release` to tag the release, create the GitHub Release, and upload the artifact.

An important benefit of this design is that release state is determined from the repository's current version, rather than by comparing the current push against the previous commit. If CI fails, subsequent pushes or workflow re-runs will therefore retry the appropriate release step.

### Testing

I tested the initial impl using basic end-to-end tests were performed on [my fork](https://github.com/MattSturgeon/nixfmt):
* [x] Normal pushes update the release-preview PR
  - https://github.com/MattSturgeon/nixfmt/actions/runs/33775807660
  - https://github.com/MattSturgeon/nixfmt/pull/3
* [x] Merged version bumps are published to GitHub Releases
  - https://github.com/MattSturgeon/nixfmt/actions/runs/33773464608
  - https://github.com/MattSturgeon/nixfmt/releases/tag/v1.4.0

I have not tested subsequent iterations on a fork yet.

### Known limitation: merge commits

The main issue we may encounter with Knope is [knope#1927](https://github.com/knope-dev/knope/issues/1927). This only affects PRs merged using a merge commit where both the PR commits and the merge commit follow conventional commits.

For that issue, we could:
- [ ] Change the [repository settings](https://github.com/NixOS/nixfmt/settings#merge-button-settings) to disable merge commits, or use the default merge commit message instead of the pull request title.
- [ ] Apply @mdaniels5757's [patches](https://github.com/knope-dev/knope/pull/1931) to a `knope` override in `main.nix`, which could then be used in both the devshell and CI.
- [ ] Live with the friction for now and manually remove duplicate changelog entries before release.

There are also several other relevant issues in [Knope's issue tracker](https://github.com/knope-dev/knope/issues), including some that I have reported while working on these migrations.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,11 +1,15 @@
-name: Release on Version Change
+name: Release
 
 on:
   push:
     branches:
       - master
-    paths:
-      - 'nixfmt.cabal'
+
+# Group workflow runs by ref.
+# New runs cancel any pending run.
+# In-progress runs are allowed to complete.
+concurrency:
+  group: push-${{ github.ref }}
 
 defaults:
   run:
@@ -14,95 +18,187 @@ defaults:
 permissions: {}
 
 jobs:
-  release-if-version-changed:
-    runs-on: ubuntu-latest
+  version:
+    name: Check version
+    runs-on: ubuntu-slim
+    timeout-minutes: 5
+
     permissions:
-      # Needed to create the release
-      contents: write
+      contents: read # Needed to checkout content & get published releases
+
+    outputs:
+      version: ${{ steps.get_version.outputs.version }}
+      published: ${{ steps.check_published.outputs.published }}
+      unpublished: ${{ steps.check_published.outputs.unpublished }}
 
     steps:
-      - name: Checkout main
+      - name: Checkout metadata
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
         with:
-          fetch-depth: 2  # need previous commit for comparison
           persist-credentials: false
+          sparse-checkout-cone-mode: false
+          sparse-checkout: nixfmt.cabal
 
       - name: Get version from nixfmt.cabal
         id: get_version
         run: |
-          CABAL_FILE="nixfmt.cabal"
-
-          # Extract the version
-          VERSION=$(grep -m1 "^version:" $CABAL_FILE | awk '{print $2}')
+          VERSION=$(grep -m1 "^version:" nixfmt.cabal | awk '{print $2}')
           echo "Version found in cabal: $VERSION"
-
-          # Get previous version from last commit
-          PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}')
-          echo "Previous version: $PREV_VERSION"
-
-          # Export for later steps
           echo "version=$VERSION" >> "$GITHUB_OUTPUT"
-          echo "prev_version=$PREV_VERSION" >> "$GITHUB_OUTPUT"
+          echo "VERSION=$VERSION" >> "$GITHUB_ENV"
 
-      - name: Check if version changed
-        id: compare
+      - name: Check if version is published
+        id: check_published
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const tag = `v${process.env.VERSION}`
+            try {
+              const { data } = await github.rest.repos.getReleaseByTag({ ...context.repo, tag })
+              core.notice(`${tag} already published (${data.published_at})`)
+              core.setOutput("published", "true")
+            } catch (err) {
+              if (err.status !== 404) throw err
+              core.notice(`${tag} not published`)
+              core.setOutput("unpublished", "true")
+            }
+
+
+  create-release-pr:
+    name: Create or update release PR
+
+    if: needs.version.outputs.published
+    needs: version
+    runs-on: ubuntu-latest # cachix-action needs a full VM
+    timeout-minutes: 5
+
+    # Match nixfmt-ci app, for testing on forks
+    permissions:
+      contents: write # Needed to push the release PR branch
+      pull-requests: write # Needed to open the release PR
+
+    steps:
+      - name: Generate App Token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        if: vars.NIXFMT_CI_APP_CLIENT_ID
+        id: app-token
+        with:
+          client-id: ${{ vars.NIXFMT_CI_APP_CLIENT_ID }}
+          private-key: ${{ secrets.NIXFMT_CI_APP_PRIVATE_KEY }}
+          permission-contents: write # Needed to push the release PR branch
+          permission-pull-requests: write # Needed to open the release PR
+
+      - name: Checkout Repository
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          fetch-depth: 0 # Knope needs to find the previous tag and read conventional commits
+          fetch-tags: true
+          persist-credentials: true
+          token: ${{ steps.app-token.outputs.token || github.token }}
+
+      - name: Get GitHub App User ID
+        id: app-id
+        if: steps.app-token.outputs.app-slug
         env:
-          PREV_VERSION: ${{ steps.get_version.outputs.prev_version }}
-          VERSION: ${{ steps.get_version.outputs.version }}
-        run: |
-          if [ "$PREV_VERSION" != "$VERSION" ]; then
-            echo "version_changed=true" >> "$GITHUB_ENV"
-            echo "::notice ::✅ Version changed from $PREV_VERSION to $VERSION"
-          else
-            echo "version_changed=false" >> "$GITHUB_ENV"
-            echo "::notice ::Version did not change, skipping release"
-          fi
-      - name: Extract changelog section for current version
-        if: ${{ env.version_changed == 'true' }}
+          APP_USERNAME: ${{ steps.app-token.outputs.app-slug }}[bot]
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: echo "app-id=$(gh api "/users/$APP_USERNAME" --jq .id)" >> "$GITHUB_OUTPUT"
+
+      - name: Configure git
         env:
-          VERSION: ${{ steps.get_version.outputs.version }}
+          APP_USER_ID: ${{ steps.app-id.outputs.app-id || '41898282' }}
+          APP_USERNAME: ${{ steps.app-token.outputs.app-slug || 'github-actions' }}[bot]
         run: |
-          echo "Extracting changelog section for version $VERSION"
-
-          # Extract everything after the "## <version>" header until the next "## "
-          awk -v ver="$VERSION" '
-            $0 ~ "^##[[:space:]]+" ver {found=1; next}
-            found && /^##[[:space:]]+/ {exit}
-            found
-          ' CHANGELOG.md > RELEASE_NOTES.md
-
-          if [ ! -s RELEASE_NOTES.md ]; then
-            echo "::error ::⚠️ No changelog section found for version $VERSION"
-            echo "(Make sure you have a header like '## $VERSION -- YYYY-MM-DD')"
-            echo "- No notes will be attached to the release -"
-            echo "No changelog section found for version $VERSION." > RELEASE_NOTES.md
-            exit 1
-          fi
-
-          echo "---- Extracted Changelog ----"
-          cat RELEASE_NOTES.md
-          echo "--------------------------"
+          git config --global user.name "$APP_USERNAME"
+          git config --global user.email "$APP_USER_ID+$APP_USERNAME@users.noreply.github.com"
 
       - name: Setup Nix
-        if: ${{ env.version_changed == 'true' }}
         uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24 # v31.11.1
 
       - name: Setup Cachix cache
-        if: ${{ env.version_changed == 'true' }}
         uses: cachix/cachix-action@38b082610b782e7e93e209c35fd730d399dee866 # v17
         with:
           name: nixos-nixfmt
-          authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
+      - name: Install Knope
+        run: |
+          nix-build main.nix -A pkgs.knope
+          realpath result/bin >> "$GITHUB_PATH"
+          echo "Installed $(result/bin/knope --version)"
+          rm result
+
+      - name: Cut release
+        id: release
+        env:
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
+        run: |
+          git switch -c prepare-release
+          knope prepare-release --verbose
+          echo "version=$(knope print-version)" >> "$GITHUB_OUTPUT"
+
+      - name: Create release PR
+        if: needs.version.outputs.version != steps.release.outputs.version
+        env:
+          VERSION: ${{ steps.release.outputs.version }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
+        run: |
+          git commit -m "chore: release $VERSION"
+          git push --force --set-upstream origin prepare-release
+          knope create-release-pr --verbose
+
+
+  publish-release:
+    name: Publish ${{ needs.version.outputs.version }}
+
+    if: needs.version.outputs.unpublished
+    needs: version
+    runs-on: ubuntu-latest # use a full VM for building static binary
+    timeout-minutes: 300 # allow for a cache-miss building the static binary
+
+    # Match nixfmt-ci app, for testing on forks
+    permissions:
+      contents: write # Needed to create the release
+
+    steps:
+      - name: Generate App Token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        if: vars.NIXFMT_CI_APP_CLIENT_ID
+        id: app-token
+        with:
+          client-id: ${{ vars.NIXFMT_CI_APP_CLIENT_ID }}
+          private-key: ${{ secrets.NIXFMT_CI_APP_PRIVATE_KEY }}
+          permission-contents: write # Needed to create the release
+
+      - name: Checkout main
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+        with:
+          persist-credentials: false
+          token: ${{ steps.app-token.outputs.token || github.token }}
+
+      - name: Setup Nix
+        uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24 # v31.11.1
+
+      - name: Setup Cachix cache
+        uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
+        with:
+          name: nixos-nixfmt
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
+
+      - name: Install Knope
+        run: |
+          nix-build main.nix -A pkgs.knope
+          realpath result/bin >> "$GITHUB_PATH"
+          echo "Installed $(result/bin/knope --version)"
+          rm result
+
+      - name: Print version
+        run: echo "::notice::Publishing $(knope print-version)"
 
       - name: Build static binary
-        if: ${{ env.version_changed == 'true' }}
-        run: |
-          nix-build -A packages.nixfmt-static
+        run: nix-build -A packages.nixfmt-static
 
-      - name: Create GitHub Release
-        if: ${{ env.version_changed == 'true' }}
+      - name: Publish release
+        run: knope release
         env:
-          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          VERSION: "${{ steps.get_version.outputs.version }}"
-        run:
-          gh release create "v$VERSION" ./result/bin/nixfmt --title "v$VERSION" --notes-file RELEASE_NOTES.md
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -2,8 +2,6 @@ name: Release
 
 on:
   push:
-    branches:
-      - master
 
 # Group workflow runs by ref.
 # New runs cancel any pending run.
@@ -26,6 +24,7 @@ jobs:
     permissions:
       contents: read # Needed to checkout content & get published releases
 
+    # Note: outputs are empty on non-release branches
     outputs:
       version: ${{ steps.get_version.outputs.version }}
       published: ${{ steps.check_published.outputs.published }}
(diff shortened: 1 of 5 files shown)


.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+          git config --global user.email "$APP_USER_ID+$APP_USERNAME@users.noreply.github.com"
+
+      - name: Setup Nix
+        uses: cachix/install-nix-action@13d8dd58da0234aa297dedd986986ccb8e7f3e24 # v31.11.1
+
+      - name: Setup Cachix cache
+        uses: cachix/cachix-action@5f2d7c5294214f71b873db4b969586b980625e71 # v17
+        with:
+          name: nixos-nixfmt
+          authToken: ${{ secrets.CACHIX_AUTH_TOKEN }}
```
MattSturgeon, line 162: Seems like the cachix setup action doesn't like running on a container.

https://github.com/MattSturgeon/nixfmt/actions/runs/33774292564/job/100712164535#step:7:22

We could set the `USER` env var as they suggest, or just run this job on a full `ubuntu-latest` VM...

MattSturgeon: Setting `USER` isn't enough: https://github.com/MattSturgeon/nixfmt/actions/runs/33774960255/job/100714354901

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+    steps:
+      - name: Generate App Token
+        uses: actions/create-github-app-token@bcd2ba49218906704ab6c1aa796996da409d3eb1 # v3.2.0
+        if: vars.NIXFMT_CI_APP_CLIENT_ID
+        id: app-token
+        with:
+          client-id: ${{ vars.NIXFMT_CI_APP_CLIENT_ID }}
+          private-key: ${{ secrets.NIXFMT_CI_APP_PRIVATE_KEY }}
+          permission-contents: write # Needed to push the release PR branch
+          permission-pull-requests: write # Needed to open the release PR
```
MattSturgeon, line 129: This uses the app added by https://github.com/NixOS/org/issues/281

I did not have this on my fork, so the fallback `github.token` was used when testing there.

You can check the [vars](https://github.com/NixOS/nixfmt/settings/variables/actions)/[secrets](https://github.com/NixOS/nixfmt/settings/secrets/actions) keys in the repo settings.

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+          sparse-checkout: nixfmt.cabal
 
       - name: Get version from nixfmt.cabal
         id: get_version
         run: |
-          CABAL_FILE="nixfmt.cabal"
-
-          # Extract the version
-          VERSION=$(grep -m1 "^version:" $CABAL_FILE | awk '{print $2}')
+          VERSION=$(grep -m1 "^version:" nixfmt.cabal | awk '{print $2}')
```
jfly, line 84: As we inject `VERSION` into code below, I'd feel more comfortable if we were stricter about what characters could possibly end up in it. I believe this `grep ... | awk ...` construction would allow any characters except whitespace.

For the record: I'm not actually worried about attacks, I just think it makes the code easier to reason about.

I think this would be reasonable solution:

```bash
VERSION=$(grep --perl-regexp --only-matching '^version: *\K[0-9.]+$' nixfmt.cabal)
```

(I'd never heard of \K before, it's a neat trick! https://regex101.com/r/0XIsLK/3)

I tried sed, but couldn't find a nice way to get it to error out if nothing matched. awk was just too awk-ward.

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
 
-          # Get previous version from last commit
-          PREV_VERSION=$(git show HEAD~1:"$CABAL_FILE" | grep -m1 "^version:" | awk '{print $2}')
-          echo "Previous version: $PREV_VERSION"
+      - name: Check if version is published
+        id: check_published
+        uses: actions/github-script@3a2844b7e9c422d3c10d287c895573f7108da1b3 # v9.0.0
+        with:
+          script: |
+            const tag = `v${process.env.VERSION}`
```
jfly, line 95: uber nit, feel free to disregard: I get a little stressed when I see JS code without trailing semicolons. Is this idiomatic for github-script?

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+
+  create-release-pr:
+    name: Create release PR
+
+    if: needs.version.outputs.published
+    needs: version
+    runs-on: ubuntu-latest # cachix-action needs a full VM
+    timeout-minutes: 5
+
+    # Match nixfmt-ci app, for testing on forks
```
jfly, line 115: Sorry, but I don't understand what this comment refers to below.

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+        id: app-id
+        if: steps.app-token.outputs.app-slug
+        env:
+          APP_USERNAME: ${{ steps.app-token.outputs.app-slug }}[bot]
+          GH_TOKEN: ${{ steps.app-token.outputs.token }}
+        run: echo "app-id=$(gh api "/users/$APP_USERNAME" --jq .id)" >> "$GITHUB_OUTPUT"
+
+      - name: Configure git
+        env:
+          APP_USER_ID: ${{ steps.app-id.outputs.app-id || '41898282' }}
```
jfly, line 149: The magic string here (`'41898282'`) could use a comment.

.github/workflows/release.yml
jfly: This seems fine, but perhaps worth reporting to cachix?

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+          realpath result/bin >> "$GITHUB_PATH"
+          echo "Installed $(result/bin/knope --version)"
+          rm result
+
+      - name: Cut release
+        id: release
         env:
-          PREV_VERSION: ${{ steps.get_version.outputs.prev_version }}
-          VERSION: ${{ steps.get_version.outputs.version }}
+          GITHUB_TOKEN: ${{ steps.app-token.outputs.token || github.token }}
```
jfly, line 174: I don't understand what the fallback `|| gitub.token` is about here. Is this to enable testing on forks?

knope.toml
```diff
@@ -0,0 +1,82 @@
...
+name = "prepare-release"
+help_text = "Prepare a new release (bump version & update changelog)"
+
+[[workflows.steps]]
+type = "PrepareRelease"
+allow_empty = true # Fail silently when there's nothing to release
+
+
+[[workflows]]
+name = "create-release-pr"
```
jfly, line 40: nit: Did you consider instead naming this `upsert-release-pr` to convey the "or update" behavior?

knope.toml
```diff
@@ -0,0 +1,82 @@
...
+name = "release"
+help_text = "Publish the current release"
+
+[[workflows.steps]]
+type = "Release"
+
+
+[[workflows]]
+name = "document-change"
+help_text = "Create a new change file to be included in the next release"
```
jfly, line 70: I cannot tell where this workflow is used.

.github/workflows/release.yml
jfly: Now that I've read the whole PR, I see that we've also taught knope about a version regex. Can we not use `knope print-version` here instead?

main.nix
```diff
@@ -163,6 +163,7 @@ in
       haskellPackages.haskell-language-server
       shellcheck
       hlint
+      knope
```
jfly, line 166: AFAICT, we don't actually use this dev shell in CI. I wonder if it's more confusing to add it to the dev shell than not. Are developers ever expected to run `knope` themselves (other than when debugging things)?

jfly (commented):
This is awesome, thanks for taking this on!

Excited to learn more about knope =)

CONTRIBUTING.md
```diff
@@ -48,6 +48,11 @@ However this is not always avoidable and they occasionally need manual adjustmen
 When editing a test or adding a new one, try to put it into a separate commit as to not mix input changes with diff changes.
 Retrospectively fixing this requires advanced git rebasing skills, and using a helper tool like [lazygit](https://github.com/jesseduffield/lazygit) is strongly recommended.
 
+### Documenting changes
```
jfly, line 51: I think it would be nice to add a `## Releasing` section to the doc as well. Doesn't have to be elaborate, just a couple sentences like:

> Whenever we have unreleased commits, there should be a release PR (maintained by [Knope](.github/workflows/release.yml)). When it's time to release, just merge that PR, [CI](.github/workflows/release.yml) will handle the rest (creating a release with artifacts).

Feel free to workshop this.

jfly: It would also be nice to have a sentence or two explaining how to do a patch release of an older version. We don't actually maintain old versions, but it would be nice to know how to do it if we had to. (The situations I can imagine where we'd want to do it are stressful, so any docs we have might be really appreciated in that stressful time.)

If the architecture we have here for releases is not compatible with doing patch releases, I don't think that's a blocker, just a sentence explaining that we cannot do them would be fine.

jfly: Oh, also: what do we do if the release job is failing? I imagine if it's a transient issue, we just retry and hope for the best. If it's due to some legitimate bug in our release flow, do we just abandon that release and cut a new one? (That's fine if so, but would be nice to explain)

jfly (commented):
> ### Known limitation: merge commits
> For that issue, we could:
>
> - Change the [repository settings](https://github.com/NixOS/nixfmt/settings#merge-button-settings) to disable merge commits, or use the default merge commit message instead of the pull request title.
> - Apply @mdaniels5757's https://github.com/knope-dev/knope/pull/1931 to a knope override in main.nix, which could then be used in both the devshell and CI.

I'd vote for changing the repository settings merge commit message. But I'm also totally OK carrying @mdaniels5757's patches instead or in addition.

.github/workflows/release.yml
MattSturgeon: Probably worth reporting, yes. But it's so common for specific actions to be incompatible with "slim"[^1] runners that I generally just take it in my stride...

Usually actions are incompatible when they use docker, but in this case it seems to be something else.

[^1]: thin container; non-VM

.github/workflows/release.yml
MattSturgeon: I used the above approach because it's what we _already_ use. I didn't want to install `knope` in this job just to do a trivial version extraction, as that'd needlessly slow down this job and add a dependency on either _nix_ or the _knope-setup_ action.

I did consider using a `sed`-based approach, using the same regex as `knope`'s config, but chose not to to minimize the diff. Happy to change to an impl more consistent with the regex used in `knope print-version` if you think that's important.

`nixfmt.cabal` and the `version:` field are controlled inputs, so I don't feel strongly about handling invalid input or covering too many edge cases. As far as I can tell, the cabal format is _mostly_ whitespace deliminated (with some comment support), so other than comments, awk should handle it _fairly_ well.

.github/workflows/release.yml
MattSturgeon: I didn't give the formatting much thought here. I don't have a strong opinion on with/without semi-colons.

If I had to make a counter-argument: I do feel like including explicit semi-colons doesn't buy you as much as you might think; they don't make javascript any stricter, and have no semantic meaning unless you compress multiple statements into one line. There's even an argument for explicit semi-colons giving a _false sense of security_, as it makes javascript _look_ like a more sane language than it actually is 😀 

Happy to use whatever formatting style you prefer, especially if you can link me to the specific tool/config (e.g. a biome config) your prefer.

.github/workflows/release.yml
MattSturgeon: The `permissions` block declares the permissions available to `github.token`. We only use `github.token` in the fallback scenario (testing on a fork without the `freecam-ci` app), so this comment explains/justifies why we declare any permissions at all.

.github/workflows/release.yml
MattSturgeon: It's the `github-actions[bot]` user ID, and is very commonly used throughout github (usually without any explanation).

I assumed it was clear from context (i.e., the `APP_USERNAME` fallback, `'github-actions'`) that this `APP_USER_ID` fallback corresponds to the ID of `github-actions`?

.github/workflows/release.yml
MattSturgeon: Yes, exactly. As per https://github.com/NixOS/nixfmt/pull/457#discussion_r3926579301 and https://github.com/NixOS/nixfmt/pull/457#discussion_r3941267711 

Most forks will not configure an app (vars & secrets), so a common pattern is to only create an app token when the app client-id variable is defined. When there is no app token, we use the gha token (`github.token`) instead, which is only useful when testing on forks.

knope.toml
MattSturgeon: I didn't consider that. I based my naming on the knope step this workflow runs (`CreatePullRequest`). It does also update the PR title/description if a PR on the tracked remote branch is already open (IIUC, it doesn't force-push though).

Happy to choose a different name, but honestly `knope --help` should make it clear enough:

> Create or update a PR for the current release

I also find `upsert` to be a little jargon-y, unless you feel strongly.

knope.toml
MattSturgeon: It is only for use by developers, who wish to interactively create a `.changeset/foo.md` change file.

main.nix
MattSturgeon: Knope is first and foremost a developer tool; devs can use it to manually do everything CI does, and they can use it to do more day-to-day things like `knope print-version` or `knope document-changes`.

The CI jobs only layer on top of that; automatically running some knope workflows depending on what is pushed.

From this perspective, I see knope as part of our dev tooling, not just a CI implementation detail.

CONTRIBUTING.md
MattSturgeon: > If the architecture we have here for releases is not compatible with doing patch releases, I don't think that's a blocker, just a sentence explaining that we cannot do them would be fine.

The architecture is compatible, but the CI implementation does not handle it yet.

`knope` could be used locally to prepare releases based on older revisions (especially if we release them on a non-`master` branch, such as a `v1.4-lts` branch). There are some limitations in the current impl though:

- releases CI only runs on pushes to `master` (can be changed, but out of scope for the initial impl)
  - So both release-preview PRs and (more importantly) auto-publish will not trigger on other branches.
- `knope create-release-pr` is hard-coded to use `master` as the target base-branch
  (but you can create a PR manually; `gh pr create --base foo`)

> Oh, also: what do we do if the release job is failing?

If it is a transient issue, we can manually re-run the workflow.

Otherwise, every push to master where the current `version` is "unpublished" will attempt to publish. For example, if we merge a "release 2.0.0" PR which fails to publish, the next PR merged ("fix broken release") will try again to publish.


> I think it would be nice to add a `## Releasing` section to the doc as well.


I agree. Happy to take a look when I get time (if not, remind me).

knope.toml
jfly: Totally fine!

CONTRIBUTING.md
```diff
@@ -48,6 +48,11 @@ However this is not always avoidable and they occasionally need manual adjustmen
...
+
+User-facing changes should be documented for inclusion in the changelog and release notes, either via [changesets](https://knope.tech/reference/concepts/change-file/) or [conventional commits](https://knope.tech/reference/concepts/conventional-commits/).
```
jfly, line 53:
```suggestion
User-facing changes should be documented for inclusion in the changelog and release notes, either via [changesets](https://knope.tech/reference/concepts/change-file/) (run `knope document-change`) or [conventional commits](https://knope.tech/reference/concepts/conventional-commits/).
```

knope.toml
jfly: Got it. I've added a suggestion to our `CONTRIBUTING.md` to document this.

.github/workflows/release.yml
jfly: It disappoints me that even with nix in our toolbelts, we are reluctant to use it to get our hands on the right tool for the job (in this case, `knope`). I don't have a complete enough mental model to engage with the performance argument: aren't we going to go on to install both nix *and* knope later on?

That all said, I'm not going to lose any sleep with the current implementation, so I'm marking this resolved.

.github/workflows/release.yml
jfly: I am firmly against arguing about formatting in PRs without tools to enforce them. Please forgive me for the comment ;)

.github/workflows/release.yml
jfly: Ah, now I see the parallelism between this line and the next. Sounds good!

.github/workflows/release.yml
jfly: I was reacting to the fact that cachix seems to have made some effort to support containers, as [your failed run](https://github.com/MattSturgeon/nixfmt/actions/runs/33774292564/job/100712164535#step:7:22) printed out a message specifically about containers:

> $USER must be set. If running in a container, try setting USER=root.::error::Action failed with error: Error: The process '/home/runner/.nix-profile/bin/cachix' failed with exit code 1

But no stress. Marking this resolved.

main.nix
jfly: Sounds good, especially now that I know about the `knope document-changes` command.

.github/workflows/release.yml
```diff
@@ -14,95 +18,187 @@ defaults:
...
+              core.setOutput("published", "true")
+            } catch (err) {
+              if (err.status !== 404) throw err
+              core.notice(`${tag} not published`)
+              core.setOutput("unpublished", "true")
+            }
+
+
+  create-release-pr:
+    name: Create release PR
```
jfly: As this was not clear to me on my first readthrough:

```suggestion
    name: Create or update release PR
```

CONTRIBUTING.md
jfly: > Otherwise, every push to master where the current version is "unpublished" will attempt to publish. For example, if we merge a "release 2.0.0" PR which fails to publish, the next PR merged ("fix broken release") will try again to publish.

Got it. I had misread the code. This sounds reasonable, but does it mean we could end up releasing code that never ends up in the changelog? Consider the following sequence of commits on main:

1. Merge a release PR. (CI fails due to some non-transient bug.)
2. Merge a PR that adds a new feature. (CI fails due to that same non-transient release bug.)
3. Merge a PR that fixes the broken release. (our `publish-release` job runs `knope release`, which IIUC, will create a github release with the correct changelog, but not actually update our CHANGELOG.md? Or it will update it, but we never commit that change? I'm trying to read https://knope.tech/reference/config-file/steps/release/, but it links to other things I haven't read, and it's time for baby's bedtime)

.github/workflows/release.yml
MattSturgeon: Sadly this is a limitation of GitHub Actions architecture. Because we have three separate jobs, they each run in their own environment. Having the first job use a `ubuntu-slim` runner also means it doesn't queue for a full VM and should spin up almost instantly.

We _could_ spend a few seconds installing nix, a few more seconds installing knope, and then use knope. But for something as trivial as reading a `version: foo` line from a text file, this doesn't seem worth it to me.

Yes, we will install nix & knope later, in the next job. But that is running in its own VM runner and doesn't share any environment with the earlier job.

The ideal world would be if GHA runners came with nix pre-installed, then it'd be a much easier decision!

jfly: Got it, thank you for the explanation! I can understand not wanting to install every tool under the sun, but it sure would be nice to be able to bring your own runner image, I guess.

CONTRIBUTING.md
MattSturgeon: Yeah, that's totally valid. If we revert from a stateful model "is this published yet" back to an event-based model "did this change the version" then we avoid those edge cases, but lose the "automatic" retry.

It really comes down to preference which model we prefer; if we want to avoid accidentally releasing with missing changelog entries after a failed release, we can switch back to event-based version bump detection.

The options are:
1. Check if the current version is published (using GH Releases API)
2. Check if the push event changed the version (e.g. compare `head_sha` with `base_sha`)
3. Check the `head_commit`'s message summary.
   can be done directly in GHA workflow expressions, without a separate job `startsWith(github.event.head_commit.message, 'chore: release')`
   but is fairly fragile; relies on release being the HEAD commit and having the expected commit message

.github/workflows/release.yml
MattSturgeon: I try to keep workflow & job names fairly short as they appear in GHA UIs. This isn't too bad though, so I'll go with it.

CONTRIBUTING.md
jfly: I like the stateful model you've created. If CI is red (or CI is just down due to an outage), I'd prefer to fix it (or wait for github to fix it) and know that all I need to worry about is getting `main` green again. In the event-based model, it's not enough to just get `main` green again, you need to reason through all the red commits and figure out if any of them were trying to do something important.

Do you know what actually happens in the scenario I described? (sorry, too busy to check myself). Does it result in an uncommitted change to our `CHANGELOG.md` that disappears when the CI runner goes away? Could we get `knope release` to error out if our CHANGELOG.md is not up to date (ideally with instructions to fix it)? That would force a human to think about if we really want to pull these intermediate changes into the release, or do something different.

I also acknowledge that I'm getting fixated on an edge case, and would be fine landing this as is.

MattSturgeon: > Merge a PR that fixes the broken release. (our publish-release job runs knope release, which IIUC, will create a github release with the correct changelog, but not actually update our CHANGELOG.md? Or it will update it, but we never commit that change? 


That's right.

All changes happen in "prepare release"; that's when the version is bumped and the changelog section is written. Sometimes I also call this "cutting" the release, "committing" the release, or "bumping" the version. This is the bit we see in the "release preview" PR.

The actual "release" (or I prefer to say "publish releases") does not make changes or commit anything; it just publishes a GitHub Release for the current commit. Our CI does this automatically when a "prepared" release is pushed to `master`.

If the changelog section is missing, it should error. But I doubt it has good instructions, as the changelog section should never be missing if the release was "prepared" by Knope.

jfly: > If the changelog section is missing, it should error.

IIUC, in the scenario I'm describing, we'd have a changelog section, but it would be "out of date".

MattSturgeon: > Could we get knope release to error out if our CHANGELOG.md is not up to date

Ah, I see what you mean. Another way to phrase that is "could we make `knope release` error if `knope prepare-release` would do something.

Theoretically, yes that's possible. But it'd need some shell code, either in Knope workflow steps, or in CI steps.

The simplest approach would be in CI: add a step that runs `knope prepare-release`, check `git diff --staged`, and error if anything was staged to be committed. Or error if `knope print-version` differs from the initial version. Either way, error if `knope prepare-release` _did something_.

The problem would be recovering from the error; once a conventional commits is in history, it can't be modified. So if we error on Knope finding post-version-bump changes in commit history, we can't easily recover from that.

A lighter sanity check might just check that no change files have been added to the `.changeset`, and raise an error saying to manually edit the CHANGELOG instead?

jfly: > Another way to phrase that is "could we make knope release error if knope prepare-release would do something."

I did not have the knope vocabulary to phrase it this way, but yes, that sounds like exactly what I'm asking.

> Theoretically, yes that's possible. But it'd need some shell code, either in Knope workflow steps, or in CI steps.

I wonder if this is something `knope` itself could be changed to do? Either by default, or a new `knope assert-release-prepared` command/workflow?

> The problem would be recovering from the error; once a conventional commits is in history, it can't be modified. So if we error on Knope finding post-version-bump changes in commit history, we can't easily recover from that.

Wouldn't the fix be to add the missing entries to the CHANGELOG.md, or revert the undesired changes?

MattSturgeon: > Wouldn't the fix be to add the missing entries to the CHANGELOG.md, or revert the undesired changes?

Knope doesn't parse individual changes _from_ a changelog; it extracts the entire _section_ as-is for release-notes (when publishing a release) or writes individual changes[^1] to a new section (when preparing a release).

Adding new changes to the changelog is correct, if they will make it into the already-prepared release. However, Knope will not know you've done that.

If you ran `knope prepare-release` again, it'll cut a second release and write another changelog section.

I'm also not sure if reverting conventional commits is understood by Knope, I'd need to check. But if we get to the point where we need to revert things on master that were merged after a failed release and we don't want them to be part of that release, then things have gotten very messy!

[^1]: a "change" can come from a conventional commit since the last tag, or a change file inside `.changeset`.

jfly: Ok, I think that clears up my mental model considerably. Thank you for walking me through this!

CONTRIBUTING.md
```diff
@@ -88,3 +114,123 @@ We call this process "expanding" groups, and the rendering algorithm will try to
...
+
+#### Push the LTS branch
+
+```console
+git push --set-upstream upstream lts/v1.2.x
+```
+
+#### Usage
+
+CI will now recognise that this as a release branch, because the branch name matches the `base` configured in `knope.toml`'s `CreatePullRequest` step.
```
jfly:
```suggestion
CI will now recognise that this is a release branch, because the branch name matches the `base` configured in `knope.toml`'s `CreatePullRequest` step.
```

jfly (approved):
Excellent work. Let's do this!

PR #462, opened 2026-09-10
ci(diff): do not run when irrelevant files are changed

MattSturgeon:
When a PR only touches non-code files, we do not need to spend CI resources running a Nixpkgs reformat diff.

In particular, this should avoid running the Nixpkgs diff on the Knope-managed release PR, which is typically updated every time something is merged into `master`.

### DIFF
--- a/.github/workflows/diff.yml
+++ b/.github/workflows/diff.yml
@@ -5,6 +5,17 @@ on:
   # We must ensure that no secrets are in any environment variables when running untrusted code, and that no secrets are persisted to disk and accessible to `readFile`
   # when evaluating Nix code
   pull_request_target: # zizmor: ignore[dangerous-triggers]
+    # When *all* paths are ignored, the workflow will not run.
+    # If any path is *not* ignored, the workflow will run.
+    paths-ignore:
+      - '*.md'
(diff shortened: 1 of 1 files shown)


jfly (commented):
I quite dislike code like this (it's something that's impossible to remember to update when we move things around in the future), but I won't stand in the way.

I would prefer an approach that leverages that fact that if a PR doesn't materially change the nixfmt derivation, then nothing needs to be done (I acknowledge I'm doing some hand waving here).

... oh wait, we haven't actually gotten our source filtering right for that, looks like updating the LICENSE does affect the derivation:

```console
$ nix eval .#nixfmt
«derivation /nix/store/adn758ziyf839gsg9vri6wbiiamm6wqs-nixfmt-1.4.0.drv»

$ echo hiya > LICENSE

$ nix eval .#nixfmt
warning: Git tree '/home/jeremy/src/github.com/NixOS/nixfmt' is dirty
«derivation /nix/store/19cj88sd43frsvy1l24wdk3acfpkxbg1-nixfmt-1.4.0.drv»
```

Double darn.

jfly (approved):

PR #465, opened 2026-09-19
fix: multiline string trailing comment idempotency

dyegoaurelio:
supersedes #443

### DIFF
--- a/test/diff/comment/in.nix
+++ b/test/diff/comment/in.nix
@@ -165,4 +165,9 @@
   "
 " # c
   t
+
+  # Trailing comment after multiline string, next item indented differently
+  "
+" # c
+      t2
 ]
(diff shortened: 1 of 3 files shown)


test/correct/string-trailing-comment.nix
```diff
@@ -0,0 +1,34 @@
+# A comment trailing the closing delimiter of a string is demoted to its own line
+# only when a token could be reindented onto the comment's column on the very
```
jfly: "a token" is pretty ambiguous. Is this rephrasing accurate?

```suggestion
# only when the first token of the next line could be reindented onto the comment's column on the very
```

jfly (commented):
I've read the tests, and they look reasonable to me. Would like to look over the implementation with someone else.

It feels to me like the complexity here is coming from the fact that we have this special behavior when trailing just happen to line up with the first token of the next line. I don't personally find that behavior terribly intuitive. Have we considered removing it? It seems to me like that would solve our idempotency issue with a net reduction in code, rather than an increase.

